PR #2, opened 2021-04-01
Support AWS preparing/cleanup for Submariner

mkolesnik:
Initial support for AWS cloud preparation and cleanup.

For easier review the prepare and cleanup are in different commits and can be reviewed in an iterative manner


pkg/aws/aws.go
```diff
@@ -0,0 +1,93 @@
...
+)
+
+const (
+	messageRetrieveVPCID      = "Retrieving VPC ID"
+	messageRetrievedVPCID     = "Retrieved VPC ID %s"
+	messagePortOpen           = "Opening port %v protocol %s for intra-cluster communications"
+	messageCreateGatewaySG    = "Creating Submariner gateway security group"
+	messageCreatedGatewaySG   = "Created Submariner gateway security group %s"
+	messageAdjustPublicSubnet = "Adjusting public subnet %s to support Submariner"
+	messageDeployGatewayNode  = "Deploying gateway node"
```
tpantelis: Most if not all of these constants are only used in one place. In that case there's no need to define a constant. I prefer message strings inlined when reading code, particularly ones with format args.

pkg/aws/aws.go
```diff
@@ -0,0 +1,93 @@
...
+type awsCloud struct {
+	client         ec2iface.EC2API
+	k8sConfig      *rest.Config
+	gwInstanceType string
+	infraID        string
+	region         string
+}
+
+// NewAwsCloud creates a new api.Cloud instance which can prepare AWS for Submariner to be deployed on it
+func NewAwsCloud(k8sConfig *rest.Config, awsSession *session.Session, infraID, gwInstanceType string) (api.Cloud, error) {
```
tpantelis: Avoids "stutter" since `aws` is the package name:

```suggestion
func NewCloud(k8sConfig *rest.Config, awsSession *session.Session, infraID, gwInstanceType string) (api.Cloud, error) {
```

pkg/aws/machinesets.go
```diff
@@ -0,0 +1,114 @@
...
+		return "", err
+	}
+
+	if len(result.Reservations) == 0 {
+		return "", newNotFoundError("reservations")
+	}
+	if len(result.Reservations[0].Instances) == 0 {
+		return "", newNotFoundError("worker instances")
+	}
+	return *result.Reservations[0].Instances[0].ImageId, nil
```
tpantelis, line 49: Do we always expect `ImageId` be non-nil?

pkg/aws/aws.go
```diff
@@ -0,0 +1,93 @@
...
+
+func (ac *awsCloud) PrepareForSubmariner(input api.PrepareForSubmarinerInput, reporter api.Reporter) error {
+	reporter.Started(messageRetrieveVPCID)
+	vpcID, err := ac.getVpcID()
+	if err != nil {
+		reporter.Failed(err)
+		return err
+	}
+	reporter.Succeeded(messageRetrievedVPCID, vpcID)
+
```
tpantelis, line 61: There's no usage of this API in this PR so I'm a bit unclear about the `Reporter` aspect. Since implementations won't be aware of the semantics of each formatted message, I assume the intent is to just log them somewhere?

tpantelis (dismissed):

pkg/aws/aws.go
mkolesnik: We use a CLI reporting one but ACM log this to some cloud service, hence I created a simple API which we can expose that should satisfy both cases.

pkg/aws/machinesets.go
mkolesnik: Yes otherwise we cant provision.

For the sake of clarity I'll add an explicit check for it.

pkg/aws/aws.go
```diff
@@ -0,0 +1,132 @@
...
+type awsCloud struct {
+	client         ec2iface.EC2API
+	k8sConfig      *rest.Config
+	gwInstanceType string
+	infraID        string
+	region         string
+}
+
+// NewCloud creates a new api.Cloud instance which can prepare AWS for Submariner to be deployed on it
+func NewCloud(k8sConfig *rest.Config, client ec2iface.EC2API, infraID, region, gwInstanceType string) api.Cloud {
```
skeeey: For ACM, we deploy gateway with `manifestwork` from the hub (broker) cluster, we don't have managed cluster kubeconfig

could we consider to provide a callback function instead of giving a kubeconfig directly, e.g. we may define a function
```
type DeployGatewayFunc func(machineSet *unstructured.Unstructured) error
type DeleteGatewayFunc func(machineSet *unstructured.Unstructured) error

type awsCloud struct {
   deployGatewayFunc DeployGatewayFunc
  deleteGatewayFunc DeleteGatewayFunc
  //...
}

func NewCloud(client ec2iface.EC2API, 
     infraID, region, gwInstanceType string, 
    deployGatewayFunc DeployGatewayFunc,
    deleteGatewayFunc DeleteGatewayFunc) api.Cloud {
//...
}
```

then in `deployGateway` function, we callback the functions to create/delete gateway

```
func (ac *awsCloud) deployGateway(vpcID, gatewaySecurityGroup string, publicSubnet *ec2.Subnet) error {
	amiID, err := ac.findAMIID(vpcID)
	if err != nil {
		return err
	}

	machineSet, err := ac.initMachineSet(vpcID, gatewaySecurityGroup, amiID, publicSubnet)
	if err != nil {
		return err
	}


	return c.deployGatewayFunc(machineSet)
}
```

thus we allow different invoker to implement its logic with its context

pkg/aws/aws.go
```diff
@@ -0,0 +1,132 @@
...
+		reporter.Succeeded("")
+	}
+
+	reporter.Started("Deploying gateway node")
+	err = ac.deployGateway(vpcID, gatewaySG, publicSubnet)
+	if err != nil {
+		reporter.Failed(err)
+		return err
+	}
+	reporter.Succeeded("")
```
skeeey: do we need to give some successful messages for this?

go.mod
```diff
@@ -1,3 +1,19 @@
...
 go 1.13
+
+require (
+	github.com/aws/aws-sdk-go v1.38.9
+	github.com/submariner-io/admiral v0.9.0-m1.0.20210303181719-5bb1de3368b5
+	k8s.io/apimachinery v0.20.5
+	k8s.io/client-go v12.0.0+incompatible
+)
+
+// Pinned to kubernetes-1.17.0
```
qiujian16: any reason we need to pin to 1.17?

pkg/api/api.go
```diff
@@ -0,0 +1,40 @@
...
+// Reporter is responsible for reporting back on the progress of the cloud preparation
+type Reporter interface {
+	// Started will report that an operation started on the cloud
+	Started(message string, args ...interface{})
+
+	// Succeeded will report that the last operation on the cloud has succeded
+	Succeeded(message string, args ...interface{})
+
+	// Failed will report that the last operation on the cloud has failed
+	Failed(err ...error)
```
qiujian16: errs I think?

pkg/api/api.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+type PortSpec struct {
+	Port     uint16
+	Protocol string
+}
+
+type PrepareForSubmarinerInput struct {
+	InternalPorts []PortSpec
```
qiujian16, line 31: Wold you add some comments here on what InternalPorts/PublicPorts are?

pkg/aws/aws.go
qiujian16: How about we could have interfaces for gw and port setup separately and a factory mode such as:

```go

type Gateway interface {
    SetupGateway() error
    CleanGateway() error
}

type Port interface {
    SetupPort(input PrepareForSubmarinerInput) error
    CleanupPort(input PrepareForSubmarinerInput) error
}

factory.WithGatewayImpl(gateway).WithPortImpl(port).WithReporter(reporter).ToCloud()
```

So in subctl case, we could have awsCloud to implement both Gateway and Port interface. And in submariner-addon case we could have a reimplementation of gateway interface?

pkg/aws/aws.go
```diff
@@ -0,0 +1,132 @@
...
+
+	reporter.Started("Creating Submariner gateway security group")
+	gatewaySG, err := ac.createGatewaySG(vpcID, input.PublicPorts)
+	if err != nil {
+		reporter.Failed(err)
+		return err
+	}
+	reporter.Succeeded("Created Submariner gateway security group %s", gatewaySG)
+
+	publicSubnet, err := ac.getTaggedPublicSubnet(vpcID)
```
zhiweiyin318, line 80: should handle the err?

pkg/aws/aws.go
```diff
@@ -0,0 +1,132 @@
...
+func (ac *awsCloud) CleanupAfterSubmariner(reporter api.Reporter) error {
+	reporter.Started(messageRetrieveVPCID)
+	vpcID, err := ac.getVpcID()
+	if err != nil {
+		reporter.Failed(err)
+		return err
+	}
+	reporter.Succeeded(messageRetrievedVPCID, vpcID)
+
+	publicSubnet, err := ac.getTaggedPublicSubnet(vpcID)
```
zhiweiyin318, line 119: ditto. should handle the err?

go.mod
mangelajo: I think it's because we pin to that in submariner-operator, but may be when being consumed, the pins from the parent project would apply so we don't need to pin here?

pkg/api/api.go
mangelajo: An alternative to this could be using channels (Accepting messages for Started,Succeeded, Failed) ? 

That would be more go-native, and easier to consume on the caller.

mangelajo: We could also keep this and provide a "channel reporter" if we think that would be useful.

go.mod
mkolesnik: I took it from the operator repo but yeah perhaps it doesn't make sense here.
I can remove the pin here and we'll see how it works i guess :man_shrugging:

pkg/api/api.go
mkolesnik: I think using args is easy and simplistic, I don't mind to change the name to `errs`

pkg/api/api.go
mkolesnik: Done

pkg/aws/aws.go
mkolesnik: Is it a requirement to deploy using a manifest work for ACM? Can't you deploy machine sets directly?

I see you do have access to a kubeConfig [1] that you use to create a work client which eventually is used by your cloud-prepare, so you could theoretically send it here instead of creating the work client.

If it's not a hard requirement, I prefer to expose as little implementation details as possible since otherwise it creates a leaky abstraction which is a bad pattern.

[1] https://github.com/open-cluster-management/submariner-addon/blob/2f122f946a38a106ee9c5f388da752c9c6c63f03/pkg/hub/manager.go#L32

pkg/aws/aws.go
mkolesnik: @tpantelis you can see the usage in this PR https://github.com/submariner-io/submariner-operator/pull/1225

pkg/aws/aws.go
qiujian16: we could for aws, but in some cases (e.g. vsphere on premise) we cannot access the spoke's apiserver. That is the reason we have a common way  to use manifeswork to deploy.

And since we already deploy the agent for manifeswork in spoke, so it is pretty natural that we also use manifestwork to create machineset.

skeeey: yeah, the manifest work is required for us, we cannot deploy machine sets directly

The submariner-addon is running on the hub cluster, The kubeConfig [1] is the hub cluster kubeconfig, we cannot use it to deploy machine sets on managed cluster.

For our process of deploying the gateway, we define the machine sets in the manifest work and add the manifest work to managed cluster namespace on the hub cluster,  then the managed clusters pull the manifest work from hub and apply the machine sets from the manifest work,  so for submariner-addon, it is running on the hub cluster, it does not have the managed cluster hub kubeconfig

mkolesnik: > yeah, the manifest work is required for us, we cannot deploy machine sets directly
> 
> The submariner-addon is running on the hub cluster, The kubeConfig [1] is the hub cluster kubeconfig, we cannot use it to deploy machine sets on managed cluster.
> 
> For our process of deploying the gateway, we define the machine sets in the manifest work and add the manifest work to managed cluster namespace on the hub cluster, then the managed clusters pull the manifest work from hub and apply the machine sets from the manifest work, so for submariner-addon, it is running on the hub cluster, it does not have the managed cluster hub kubeconfig

OK thanks for the detailed explanation, I'll make the necessary adjustments

pkg/aws/aws.go
```diff
@@ -12,19 +13,37 @@ const (
 	messageRetrievedVPCID = "Retrieved VPC ID %s"
 )
 
+// MachineSetDeployer can deploy and delete machinesets from OCP
+type MachineSetDeployer interface {
```
mkolesnik, line 17: @qiujian16 @skeeey let me know what you think of this approach?

I think it matches your request but I want to be sure.

pkg/aws/aws.go
mkolesnik: > How about we could have interfaces for gw and port setup separately and a factory mode such as:
> 
> ```go
> type Gateway interface {
>     SetupGateway() error
>     CleanGateway() error
> }

Please see the latest commit and let me know if it looks good to you to use.

> 
> type Port interface {
>     SetupPort(input PrepareForSubmarinerInput) error
>     CleanupPort(input PrepareForSubmarinerInput) error
> }

For port setup I'm not sure why'd you need a separate functionality?
It seems that my implementation here pretty much does the same port setup that you are doing and the input allows to easily differentiate between the internal and external ports to open.
Let me know if I misunderstand.

>

pkg/aws/aws.go
qiujian16: thanks, I think it should work. But it seems that it tightly binds to machineset api? Will we want to use the same interface to support non-ocp env in the future?

skeeey: How about use the machine information instead of machineset api?
```
type MachineConfig struct {
	InfraId           string
	AZ                string
	AMIId             string
	InstanceType      string
	Region            string
	SecurityGroupName string
	SubnetName        string
}
```

mkolesnik: > thanks, I think it should work. But it seems that it tightly binds to machineset api? Will we want to use the same interface to support non-ocp env in the future?

I figure for now we know for sure we will be deploying an extra node on AWS, but for GCP, azure and others its not clear if that will be mandatory or will use another method, so seems quite logical top put this in AWS and not in the general API.

We also have a POC to do regarding putting the gateway behind a standard k8s service, so the extra gateway node might not be a requirement in the long run,



> How about use the machine information instead of machineset api?
> 
> ```
> type MachineConfig struct {
> 	InfraId           string
> 	AZ                string
> 	AMIId             string
> 	InstanceType      string
> 	Region            string
> 	SecurityGroupName string
> 	SubnetName        string
> }
> ```

In this case you'll have to hold your own copy of the machineset yaml, and seeing as this is more of an "implementation detail" rather than API we want to expose, I'm not sure there's a real benefit to this proposal.

mkolesnik: Also I agree this is quite OCP focused but the problem as I see it is the more we take outside the internal implementation, the more we leak the internal details outside instead of having a nice API where the caller is abstracted from the internal details which might change.

qiujian16: I was thinking that we could have an OCPAWSGatewayDeployerImpl, and in NewOCPAwsGatewayDeployerImpl(...) func we put all the related info to construct machineset. So when calling Deployer(), we will not need machinset specific configuration.

But since we plan to avoid gateway deployment, and your concern makes sense...We could live it as the first step and see how its moving forward :) @skeeey would you double confirm that current API abstraction could be also implemented by us?

skeeey: yeah, as the fist step I think this is fine,  we can use this

pkg/aws/subnets.go
```diff
@@ -0,0 +1,67 @@
...
+	result, err := ac.client.DescribeSubnets(&ec2.DescribeSubnetsInput{Filters: filters})
+	if err != nil {
+		return nil, err
+	}
+
+	if len(result.Subnets) == 0 {
+		return nil, newNotFoundError("public subnet")
+	}
+
+	return result.Subnets[0], nil
```
skeeey, line 26: I think we need to check whether we can create a machine with the specified machine type in this subnet, we have encountered an issue that the returned subnet does not support the machine type that we specified.

you may use `DescribeInstanceTypeOfferings` api to check, there is an example: https://github.com/open-cluster-management/submariner-addon/blob/main/pkg/cloud/aws/aws.go#L348

mkolesnik: I'm looking into it already, thanks for noticing :)

mkolesnik: I've addressed this in #5

mkolesnik:
It seems that there's generally a consensus on the API, and follow up PRs address some other issues and concerns.

I suggest that, if we're all fine with the API, we can start merging the PRs and address any further issues that might pop up in follow ups?

/cc @skeeey @zhiweiyin318 @qiujian16

tpantelis:
> I suggest that, if we're all fine with the API, we can start merging the PRs and address any further issues that might pop up in follow ups?
> 

that would be good then we can start adding linting and other GHAs.

mangelajo (approved):

mangelajo:
Based on the comments and code I +1'ed it, @qiujian16 @skeeey does github let you vote?

qiujian16:
+1 from me, thanks!

tpantelis (approved):

PR #6, opened 2021-04-22
Add linting github action

tpantelis:


tpantelis:
hmm... the linting workflow didn't run. @dfarrell07 is this a chicken-and-the-egg thing, ie the file must be added first then will work on subsequent PRs, or did I miss something? What's odd is that the job added by https://github.com/submariner-io/cloud-prepare/pull/7 ran.

dfarrell07:
It should run in this PR, it shouldn't be a chicken-an-egg situation. I wonder if there's a subtle error in the GHA definitions? That causes them to silently not run, as we're seeing here. I could also imagine it just needing a force push or something for GH to pick up the new jobs and run them.

tpantelis:
> It should run in this PR, it shouldn't be a chicken-an-egg situation. I wonder if there's a subtle error in the GHA definitions? That causes them to silently not run, as we're seeing here. I could also imagine it just needing a force push or something for GH to pick up the new jobs and run them.

I'll try that.  I copied this from submariner as is (only removed the code-gen job) so it should be fine.

Tried it - still doesn't run. I'll try paring it down to the simplest job. The issue may be one specific  job.

dfarrell07:
Well, I'm not sure what you did but I see it's working now. :heavy_check_mark:

dfarrell07 (approved):

pkg/aws/gw-machineset.yaml
```diff
@@ -4,7 +4,7 @@ kind: MachineSet
 metadata:
   labels:
     machine.openshift.io/cluster-api-cluster: {{.InfraID}}
-  name: {{.InfraID}}-submariner-gw-{{.AZ}}
```
mkolesnik: This commented out on purpose?

mkolesnik (commented):
Thanks Tom, one question :)

pkg/aws/gw-machineset.yaml
tpantelis: crap - I didn't mean that - thanks for catching it.  The yaml linter was failing on that line due to the "{{*}}" part so I was playing around with it.  I think it's a false positive and I ended up taking out that linter.

mangelajo (approved):

skitt (approved):

PR #9, opened 2021-04-23
Add branch check GHA

tpantelis:


dfarrell07 (approved):

mangelajo (changes requested):
There is another version we have that also matches to release-xxxx. We should use the other one.

tpantelis:
> There is another version we have that also matches to release-xxxx. We should use the other one.

I copied this from the latest `submariner` project. It's also the same in the other projects except `submariner-operator` which has no such job.

mkolesnik (approved):

mangelajo (approved):

PR #86, opened 2021-08-20
Add YAML linting, fix errors

dfarrell07:


dfarrell07:
There's one YAML linting failure remaining.

```
[linting]$ make yamllint
yamllint .
./pkg/aws/gw-machineset.yaml
  7:21      error    syntax error: expected <block end>, but found '<scalar>' (syntax)
```

I think the issue is that this isn't actually a YAML file, it's a template. It's also not actually used currently.

```
  // TODO: Not working properly, but we should revisit this as it makes more sense
  // tpl, err := template.ParseFiles("pkg/aws/gw-machineset.yaml")
```

tpantelis (approved):

dfarrell07:
So to summarize, this is blocked on figuring out what we want to do about the non-YAML (templated) .yaml file described above.

skitt:
Depends on #95.

skitt (approved):

PR #92, opened 2021-08-26
GCP support for deploying dedicated Gateway nodes

### LINKED ISSUE #91: cloud-prepare: add support for deploying new GW nodes on GCP
**What would you like to be added**:
We need support for deploying new Dedicated Gateway nodes on the GCP Cluster instead of choosing an existing one from the Worker node for certain scenarios.

sridhargaddam:
This PR implements the following
1. Supports deploying new Gateway nodes using the OCP machineset API
2. Cleanup of any existing gateway nodes and firewall rules
3. Fixes the public-firewall-rules to be applied only to gateway nodes

Fixes issue: https://github.com/submariner-io/cloud-prepare/issues/91

### DIFF
--- a/pkg/gcp/client/client.go
+++ b/pkg/gcp/client/client.go
@@ -19,7 +19,9 @@ package client
 
 import (
 	"context"
+	"fmt"
 	"net/http"
+	"strings"
 
 	compute "google.golang.org/api/compute/v1"
 	"google.golang.org/api/googleapi"
@@ -34,9 +36,16 @@ type Interface interface {
 	GetFirewallRule(projectID, name string) (*compute.Firewall, error)
 	DeleteFirewallRule(projectID, name string) error
 	UpdateFirewallRule(projectID, name string, rule *compute.Firewall) error
+	GetInstance(zone string, instance string) (*compute.Instance, error)
+	ListInstances(zone string) (*compute.InstanceList, error)
+	ListZones() (*compute.ZoneList, error)
+	InstanceHasPublicIP(instance *compute.Instance) (bool, error)
+	ConfigurePublicIPOnInstance(instance *compute.Instance) error
+	DeletePublicIPOnInstance(instance *compute.Instance) error
 }
 
 type gcpClient struct {
+	projectID     string
 	computeClient *compute.Service
 }
 
@@ -59,7 +68,7 @@ func (g *gcpClient) UpdateFirewallRule(projectID, name string, rule *compute.Fir
 	return err
 }
 
-func NewClient(options []option.ClientOption) (Interface, error) {
+func NewClient(projectID string, options []option.ClientOption) (Interface, error) {
 	ctx := context.TODO()
 
 	computeClient, err := compute.NewService(ctx, options...)
@@ -68,6 +77,7 @@ func NewClient(options []option.ClientOption) (Interface, error) {
 	}
 
 	return &gcpClient{
+		projectID:     projectID,
 		computeClient: computeClient,
 	}, nil
 }
@@ -79,4 +89,83 @@ func IsGCPNotFoundError(err error) bool {
 	}
 
 	return gerr.Code == http.StatusNotFound
+}
+
+func (g *gcpClient) GetInstance(zone, instance string) (*compute.Instance, error) {
+	resp, err := g.computeClient.Instances.Get(g.projectID, zone, instance).Context(context.TODO()).Do()
+	if err != nil {
+		return nil, err
+	}
+
+	return resp, nil
+}
+
+func (g *gcpClient) ListInstances(zone string) (*compute.InstanceList, error) {
+	resp, err := g.computeClient.Instances.List(g.projectID, zone).Context(context.TODO()).Do()
+	if err != nil {
+		return nil, err
+	}
+
+	return resp, nil
+}
+
+func (g *gcpClient) ListZones() (*compute.ZoneList, error) {
+	resp, err := g.computeClient.Zones.List(g.projectID).Context(context.TODO()).Do()
+	if err != nil {
+		return nil, err
+	}
+
+	return resp, nil
+}
+
+func (g *gcpClient) InstanceHasPublicIP(instance *compute.Instance) (bool, error) {
+	if len(instance.NetworkInterfaces) == 0 {
+		return false, fmt.Errorf("there are no network interfaces for instance %s", instance.Name)
+	}
+
+	networkInterface := instance.NetworkInterfaces[0]
+	if len(networkInterface.AccessConfigs) > 0 {
+		return true, nil
+	}
+
+	return false, nil
+}
+
+func (g *gcpClient) ConfigurePublicIPOnInstance(instance *compute.Instance) error {
+	if len(instance.NetworkInterfaces) == 0 {
+		return fmt.Errorf("there are no network interfaces for instance %s", instance.Name)
+	}
+
+	// The zone of an instance is on URL, so we just need the latest value
+	zone := instance.Zone[strings.LastIndex(instance.Zone, "/")+1:]
+	networkInterface := instance.NetworkInterfaces[0]
+	// Public IP has already been enabled for this instance
+	if len(networkInterface.AccessConfigs) > 0 {
+		return nil
+	}
+
+	if _, err := g.computeClient.Instances.AddAccessConfig(
+		g.projectID, zone, instance.Name, networkInterface.Name, &compute.AccessConfig{}).
+		Context(context.TODO()).Do(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (g *gcpClient) DeletePublicIPOnInstance(instance *compute.Instance) error {
+	if len(instance.NetworkInterfaces) == 0 {
+		return fmt.Errorf("there are no network interfaces for instance %s", instance.Name)
+	}
+
+	// The zone of an instance is on URL, so we just need the latest value
+	zone := instance.Zone[strings.LastIndex(instance.Zone, "/")+1:]
+	networkInterface := instance.NetworkInterfaces[0]
+	if _, err := g.computeClient.Instances.DeleteAccessConfig(
+		g.projectID, zone, instance.Name, "External NAT", networkInterface.Name).
+		Context(context.TODO()).Do(); err != nil {
+		return err
+	}
+
+	return nil
 }
--- a/pkg/gcp/client/client.go
+++ b/pkg/gcp/client/client.go
@@ -92,30 +92,15 @@ func IsGCPNotFoundError(err error) bool {
 }
 
 func (g *gcpClient) GetInstance(zone, instance string) (*compute.Instance, error) {
-	resp, err := g.computeClient.Instances.Get(g.projectID, zone, instance).Context(context.TODO()).Do()
-	if err != nil {
-		return nil, err
-	}
-
-	return resp, nil
+	return g.computeClient.Instances.Get(g.projectID, zone, instance).Context(context.TODO()).Do()
 }
 
 func (g *gcpClient) ListInstances(zone string) (*compute.InstanceList, error) {
-	resp, err := g.computeClient.Instances.List(g.projectID, zone).Context(context.TODO()).Do()
-	if err != nil {
-		return nil, err
-	}
-
-	return resp, nil
+	return g.computeClient.Instances.List(g.projectID, zone).Context(context.TODO()).Do()
 }
 
 func (g *gcpClient) ListZones() (*compute.ZoneList, error) {
-	resp, err := g.computeClient.Zones.List(g.projectID).Context(context.TODO()).Do()
-	if err != nil {
-		return nil, err
-	}
-
-	return resp, nil
+	return g.computeClient.Zones.List(g.projectID).Context(context.TODO()).Do()
 }
 
 func (g *gcpClient) InstanceHasPublicIP(instance *compute.Instance) (bool, error) {
@@ -124,11 +109,8 @@ func (g *gcpClient) InstanceHasPublicIP(instance *compute.Instance) (bool, error
 	}
 
 	networkInterface := instance.NetworkInterfaces[0]
-	if len(networkInterface.AccessConfigs) > 0 {
-		return true, nil
-	}
 
-	return false, nil
+	return len(networkInterface.AccessConfigs) > 0, nil
 }
 
 func (g *gcpClient) ConfigurePublicIPOnInstance(instance *compute.Instance) error {
@@ -144,13 +126,11 @@ func (g *gcpClient) ConfigurePublicIPOnInstance(instance *compute.Instance) erro
 		return nil
 	}
 
-	if _, err := g.computeClient.Instances.AddAccessConfig(
-		g.projectID, zone, instance.Name, networkInterface.Name, &compute.AccessConfig{}).
-		Context(context.TODO()).Do(); err != nil {
-		return err
-	}
+	_, err := g.computeClient.Instances.AddAccessConfig(g.projectID, zone, instance.Name,
+		networkInterface.Name, &compute.AccessConfig{}).
+		Context(context.TODO()).Do()
 
-	return nil
+	return err
 }
 
 func (g *gcpClient) DeletePublicIPOnInstance(instance *compute.Instance) error {
@@ -161,11 +141,9 @@ func (g *gcpClient) DeletePublicIPOnInstance(instance *compute.Instance) error {
 	// The zone of an instance is on URL, so we just need the latest value
 	zone := instance.Zone[strings.LastIndex(instance.Zone, "/")+1:]
 	networkInterface := instance.NetworkInterfaces[0]
-	if _, err := g.computeClient.Instances.DeleteAccessConfig(
+	_, err := g.computeClient.Instances.DeleteAccessConfig(
 		g.projectID, zone, instance.Name, "External NAT", networkInterface.Name).
-		Context(context.TODO()).Do(); err != nil {
-		return err
-	}
+		Context(context.TODO()).Do()
 
-	return nil
+	return err
 }
--- a/pkg/gcp/gcp.go
+++ b/pkg/gcp/gcp.go
@@ -29,18 +29,18 @@ import (
 )
 
 const (
-	messageRetrieveZones           = "Retrieving zones in the project"
-	messageRetrievedZones          = "Retrieved the zones"
-	messageValidateCurrentGWCount  = "Verifying if current gateways match the required number of gateways"
-	messageValidatedCurrentGWs     = "Current gateways match the required number of gateways"
-	messageDeployGatewayNode       = "Deploying gateway node"
-	messageDeployedGatewayNode     = "Successfully deployed gateway node"
-	messageDeployGatewayNodeFailed = "Could not deploy the required number of Gateway nodes"
-	messageCreateExtFWRules        = "Configuring the required firewall rules for inter-cluster traffic"
-	messageDeleteExtFWRules        = "Deleting the Submariner gateway firewall rules"
-	messageDeletedExtFWRules       = "Successfully deleted the firewall rules"
-	messageVerifyCurrentGWCount    = "Looking for current gateways in the project that need to be deleted"
-	messageVerifiedCurrentGWCount  = "Successfully deleted the gateway nodes"
+	messageRetrieveZones              = "Retrieving zones in the project"
+	messageRetrievedZones             = "Retrieved the zones"
+	messageValidateCurrentGWCount     = "Verifying if current gateways match the required number of gateways"
+	messageValidatedCurrentGWs        = "Current gateways match the required number of gateways"
+	messageDeployGatewayNode          = "Deploying gateway node"
+	messageDeployedGatewayNode        = "Successfully deployed gateway node"
+	messageInsufficientZonesForDeploy = "There are insufficient zone instances to deploy the required number of gateways"
+	messageCreateExtFWRules           = "Configuring the required firewall rules for inter-cluster traffic"
+	messageDeleteExtFWRules           = "Deleting the Submariner gateway firewall rules"
+	messageDeletedExtFWRules          = "Successfully deleted the firewall rules"
+	messageVerifyCurrentGWCount       = "Looking for current gateways in the project that need to be deleted"
+	messageVerifiedCurrentGWCount     = "Successfully deleted the gateway nodes"
 )
 
 type gcpCloud struct {
--- a/pkg/gcp/gcp.go
+++ b/pkg/gcp/gcp.go
@@ -28,17 +28,33 @@ import (
 	"google.golang.org/api/compute/v1"
 )
 
+const (
+	messageRetrieveZones          = "Retrieving zones in the project"
+	messageRetrievedZones         = "Retrieved the zones"
+	messageValidateCurrentGWCount = "Verifying if current gateways match the required number of gateways"
+	messageValidatedCurrentGWs    = "Current gateways match the required number of gateways"
+	messageDeployGatewayNode      = "Deploying gateway node"
+	messageDeployedGatewayNode    = "Successfully deployed gateway node"
+	messageCreateExtFWRules       = "Configuring the required firewall rules for inter-cluster traffic"
+	messageDeleteExtFWRules       = "Deleting the Submariner gateway firewall rules"
+	messageDeletedExtFWRules      = "Successfully deleted the firewall rules"
+	messageVerifyCurrentGWCount   = "Looking for current gateways in the project that need to be deleted"
+	messageVerifiedCurrentGWCount = "Successfully deleted the gateway nodes"
+)
+
 type gcpCloud struct {
 	infraID   string
+	region    string
 	projectID string
 	client    gcpclient.Interface
 }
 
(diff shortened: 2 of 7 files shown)


pkg/gcp/client/client.go
```diff
@@ -80,3 +90,82 @@ func IsGCPNotFoundError(err error) bool {
...
 	return gerr.Code == http.StatusNotFound
 }
+
+func (g *gcpClient) GetInstance(zone, instance string) (*compute.Instance, error) {
+	resp, err := g.computeClient.Instances.Get(g.projectID, zone, instance).Context(context.TODO()).Do()
+	if err != nil {
+		return nil, err
+	}
+
+	return resp, nil
```
tpantelis:
```suggestion
	return g.computeClient.Instances.Get(g.projectID, zone, instance).Context(context.TODO()).Do()
```

pkg/gcp/client/client.go
```diff
@@ -80,3 +90,82 @@ func IsGCPNotFoundError(err error) bool {
...
+}
+
+func (g *gcpClient) ListInstances(zone string) (*compute.InstanceList, error) {
+	resp, err := g.computeClient.Instances.List(g.projectID, zone).Context(context.TODO()).Do()
+	if err != nil {
+		return nil, err
+	}
+
+	return resp, nil
```
tpantelis:
```suggestion
	return g.computeClient.Instances.List(g.projectID, zone).Context(context.TODO()).Do()
```

pkg/gcp/client/client.go
```diff
@@ -80,3 +90,82 @@ func IsGCPNotFoundError(err error) bool {
...
+}
+
+func (g *gcpClient) ListZones() (*compute.ZoneList, error) {
+	resp, err := g.computeClient.Zones.List(g.projectID).Context(context.TODO()).Do()
+	if err != nil {
+		return nil, err
+	}
+
+	return resp, nil
```
tpantelis:
```suggestion
	return g.computeClient.Zones.List(g.projectID).Context(context.TODO()).Do()
```

pkg/gcp/client/client.go
```diff
@@ -80,3 +90,82 @@ func IsGCPNotFoundError(err error) bool {
...
+	if len(instance.NetworkInterfaces) == 0 {
+		return false, fmt.Errorf("there are no network interfaces for instance %s", instance.Name)
+	}
+
+	networkInterface := instance.NetworkInterfaces[0]
+	if len(networkInterface.AccessConfigs) > 0 {
+		return true, nil
+	}
+
+	return false, nil
```
tpantelis:
```suggestion
	return len(networkInterface.AccessConfigs) > 0, nil
```

pkg/gcp/client/client.go
```diff
@@ -80,3 +90,82 @@ func IsGCPNotFoundError(err error) bool {
...
+		return nil
+	}
+
+	if _, err := g.computeClient.Instances.AddAccessConfig(
+		g.projectID, zone, instance.Name, networkInterface.Name, &compute.AccessConfig{}).
+		Context(context.TODO()).Do(); err != nil {
+		return err
+	}
+
+	return nil
```
tpantelis:
```suggestion
	_, err := g.computeClient.Instances.AddAccessConfig(
		g.projectID, zone, instance.Name, networkInterface.Name, &compute.AccessConfig{}).
		Context(context.TODO()).Do()
	return err
```

pkg/gcp/client/client.go
```diff
@@ -80,3 +90,82 @@ func IsGCPNotFoundError(err error) bool {
...
+	// The zone of an instance is on URL, so we just need the latest value
+	zone := instance.Zone[strings.LastIndex(instance.Zone, "/")+1:]
+	networkInterface := instance.NetworkInterfaces[0]
+	if _, err := g.computeClient.Instances.DeleteAccessConfig(
+		g.projectID, zone, instance.Name, "External NAT", networkInterface.Name).
+		Context(context.TODO()).Do(); err != nil {
+		return err
+	}
+
+	return nil
```
tpantelis:
```suggestion
	_, err := g.computeClient.Instances.DeleteAccessConfig(
		g.projectID, zone, instance.Name, "External NAT", networkInterface.Name).
		Context(context.TODO()).Do()
		
	return err
```

pkg/gcp/ocpgwdeployer.go
```diff
@@ -0,0 +1,285 @@
...
+
+	reporter.Succeeded("Opened External ports %q with firewall rule %q on GCP",
+		formatPorts(input.PublicPorts), externalIngress.Name)
+
+	reporter.Started(messageRetrieveZones)
+
+	zones, err := d.gcp.client.ListZones()
+	if err != nil {
+		reporter.Failed(err)
+		return fmt.Errorf("failed to list the zones in the project %q. %v", d.gcp.projectID, err)
```
tpantelis: You augment the return error with additional context - perhaps we should do the same with the `reporter`? Applies to other places as well

```suggestion
		return reportFailure(reporter, err, "failed to list the zones in the project %q", d.gcp.projectID)
		
	...
	func reportFailure(reporter api.Reporter, failure error, format string, args ...string) {
	    err  := errors.WithMessagef(failure, format, args)
	    reporter.Failed(err)
	    
	    return err
	}
```

pkg/gcp/ocpgwdeployer.go
```diff
@@ -0,0 +1,285 @@
...
+		return nil
+	}
+
+	if len(currentGWInstanceList) < input.Gateways {
+		gatewayNodesToDeploy := input.Gateways - len(currentGWInstanceList)
+
+		reporter.Started(messageDeployGatewayNode)
+
+		for _, zone := range zones.Items {
+			err := d.deployGateway(zone.Name)
```
tpantelis: We use criteria on L84 and L101 above to ignore zones when determining if a zone has a current GW instance. But we don't use that same criteria here when selecting a zone on which to deploy nor do we ensure we deploy on a new zone, ie one we haven't already deployed on. Unless I'm missing something or my understanding of the code and concepts is incorrect....

pkg/gcp/ocpgwdeployer.go
```diff
@@ -0,0 +1,285 @@
...
+				reporter.Failed(err)
+				return err
+			}
+
+			gatewayNodesToDeploy--
+			if gatewayNodesToDeploy <= 0 {
+				reporter.Succeeded(messageDeployedGatewayNode)
+				return nil
+			}
+		}
```
tpantelis, line 135: So what if `gatewayNodesToDeploy` doesn't reach 0 and we end up here. That would mean we didn't find enough zones to satisfy the requested # of gateways. Should we return an error and/or report it?

pkg/gcp/ocpgwdeployer.go
```diff
@@ -0,0 +1,285 @@
...
+	}
+
```
tpantelis, line 144: What if `len(currentGWInstanceList) > input.Gateways`?  Should we undeploy gateways to match the request?

pkg/gcp/ocpgwdeployer.go
```diff
@@ -0,0 +1,285 @@
...
+		reporter.Failed(err)
+		return fmt.Errorf("failed to list the zones in the project %q. %v", d.gcp.projectID, err)
+	}
+
+	reporter.Succeeded(messageRetrievedZones)
+	reporter.Started(messageVerifyCurrentGWCount)
+
+	for _, zone := range zones.Items {
+		region := zone.Region[strings.LastIndex(zone.Region, "/")+1:]
+		if region != d.gcp.region {
```
tpantelis: L84 above checks `zone.Region` as is which differs with L236 here. Shouldn't they behave the same?

pkg/ocp/machinesets.go
```diff
@@ -70,6 +74,41 @@ func (msd *k8sMachineSetDeployer) clientFor(obj runtime.Object) (resource.Interf
 	return resource.ForDynamic(dynamicClient), nil
 }
 
+func (msd *k8sMachineSetDeployer) GetWorkerNodeImage(machineSet *unstructured.Unstructured, infraID string) (string, error) {
+	machineSetClient, err := msd.clientFor(machineSet)
+	if err != nil {
+		return "", err
+	}
+
+	// TODO: After implementing a ListAll method in admiral, modify this code accordingly.
```
tpantelis, line 83: Where would this method go?

tpantelis (changes requested):

pkg/gcp/ocpgwdeployer.go
sridhargaddam: Thats true. Basically, the idea is that the cloud-prepare API should be idempotent and if user calls with the same arguments any number of times the result should be the same. Currently, we do not have a use-case where users will try to increase/decrease the number of Gateways in a subsequent call. If you think I should also check the zone, I can do it in a subsequent PR. Hope its fine with you.

pkg/gcp/ocpgwdeployer.go
sridhargaddam: Added a check

pkg/gcp/ocpgwdeployer.go
sridhargaddam: Currently we do not have a use-case to reduce gateway count (except for cleanup). If there is any such requirement in future, we can look into it.

pkg/gcp/ocpgwdeployer.go
sridhargaddam: Good catch @tpantelis. You are just awesome.

sridhargaddam: Good catch @tpantelis. You are just awesome.

pkg/ocp/machinesets.go
sridhargaddam: In admiral/pkg/resource/interface.go, interface methods do not include ListAll

sridhargaddam (commented):
Thanks for reviewing the PR @tpantelis

pkg/gcp/ocpgwdeployer.go
tpantelis: > Thats true. Basically, the idea is that the cloud-prepare API should be idempotent and if user calls with the same arguments any number of times the result should be the same. 

Right and that's handled on line L110 above.

> Currently, we do not have a use-case where users will try to increase/decrease the number of Gateways in a subsequent call.

hmm... but this is handling the increase case. If there's no use case for that then we should only handle the case where the GWs are provisioned for the first time, ie only deploy GWs `if len(currentGWInstanceList) == 0`. Otherwise I think it would be inconsistent if the increase case is handled but not the decrease case. I will say in the add-on, the `submarinerConfigController` that labels nodes as GWs does handle all cases tho I'm not sure how this relates to that.

 > If you think I should also check the zone, I can do it in a subsequent PR. Hope its fine with you.

Not sure what you mean by check the zone, but it seems to me the this code should use the same zone selection criteria as above and ensure we pick one that isn't already a GW. Perhaps we should compile an array of eligible zone names in the loop above in parallel with the `currentGWInstanceList` and loop over that list here?

pkg/gcp/gcp.go
```diff
@@ -28,17 +28,34 @@ import (
...
 )
 
+const (
+	messageRetrieveZones           = "Retrieving zones in the project"
+	messageRetrievedZones          = "Retrieved the zones"
+	messageValidateCurrentGWCount  = "Verifying if current gateways match the required number of gateways"
+	messageValidatedCurrentGWs     = "Current gateways match the required number of gateways"
+	messageDeployGatewayNode       = "Deploying gateway node"
+	messageDeployedGatewayNode     = "Successfully deployed gateway node"
+	messageDeployGatewayNodeFailed = "Could not deploy the required number of Gateway nodes"
```
tpantelis:
```suggestion
	messageInsufficientInstancesForDeploy = "There are insufficient zone instances to deploy the required number of gateways"
```

pkg/gcp/ocpgwdeployer.go
```diff
@@ -0,0 +1,291 @@
...
+
+			gatewayNodesToDeploy--
+			if gatewayNodesToDeploy <= 0 {
+				reporter.Succeeded(messageDeployedGatewayNode)
+				return nil
+			}
+		}
+
+		if gatewayNodesToDeploy > 0 {
+			reporter.Succeeded(messageDeployGatewayNodeFailed)
```
tpantelis: It didn't technically succeed or may have partially succeeded. Perhaps this should this should call `Failed`? I'm not sure how the `Reporter` interface is/will be used by implementors.

tpantelis (changes requested):

pkg/gcp/ocpgwdeployer.go
sridhargaddam: Changed it to use messageDeployGatewayNodeFailed

pkg/gcp/ocpgwdeployer.go
sridhargaddam: > hmm... but this is handling the increase case. If there's no use case for that then we should only handle the case where the GWs are provisioned for the first time, ie only deploy GWs `if len(currentGWInstanceList) == 0`. Otherwise I think it would be inconsistent if the increase case is handled but not the decrease case. 

Although we don't have an immediate use-case to support increasing the gateways, it will surely be helpful when the user wants to modify the cluster from a non-HA deployment (aka single GWNode) to an HA deployment (with multiple gateway nodes) without affecting the datapath. At the same time, we do not support decreasing the gateway count because it would affect the datapath connectivity as we might accidentally delete the active Gateway node in the cluster. Also JFYI, in OpenShift AFAIK it supports increasing the POD/SVC CIDR ranges but does not allow decreasing them.

> Not sure what you mean by check the zone, but it seems to me the this code should use the same zone selection criteria as above and ensure we pick one that isn't already a GW. Perhaps we should compile an array of eligible zone names in the loop above in parallel with the `currentGWInstanceList` and loop over that list here?

Done.

sridhargaddam (commented):
Thanks for reviewing @tpantelis. Addressed the review comments.

pkg/gcp/ocpgwdeployer.go
```diff
@@ -0,0 +1,303 @@
...
+
+	zones, err := d.gcp.client.ListZones()
+	if err != nil {
+		return reportFailure(reporter, err, "failed to list the zones in the project %q. %v", d.gcp.projectID)
+	}
+
+	reporter.Succeeded(messageRetrievedZones)
+	reporter.Started(messageValidateCurrentGWCount)
+
+	zonesWithSubmarinerGW := stringset.NewSynchronized()
```
tpantelis: This doesn't need to be synchronized since it's used locally on a single thread.

pkg/gcp/ocpgwdeployer.go
```diff
@@ -0,0 +1,303 @@
...
+						}
+
+						break
+					}
+				}
+			}
+		}
+	}
+
+	if len(currentGWInstanceList) == input.Gateways {
```
tpantelis: `currentGWInstanceList` is now redundant with `zonesWithSubmarinerGW`

```suggestion
	if len(zonesWithSubmarinerGW) == input.Gateways {
```

pkg/gcp/ocpgwdeployer.go
```diff
@@ -0,0 +1,303 @@
...
+		reporter.Succeeded(messageValidatedCurrentGWs)
+		return nil
+	}
+
+	if len(currentGWInstanceList) < input.Gateways {
+		gatewayNodesToDeploy := input.Gateways - len(currentGWInstanceList)
```
tpantelis:
```suggestion
	if len(zonesWithSubmarinerGW) < input.Gateways {
		gatewayNodesToDeploy := input.Gateways - len(zonesWithSubmarinerGW)
```

pkg/gcp/ocpgwdeployer.go
```diff
@@ -0,0 +1,303 @@
...
+			}
+
+			if hasPublicIP {
+				for _, tag := range instance.Tags.Items {
+					if tag == submarinerGatewayNodeTag {
+						currentGWInstanceList = append(currentGWInstanceList, instance)
+
+						if !zonesWithSubmarinerGW.Contains(zone.Name) {
+							zonesWithSubmarinerGW.Add(zone.Name)
+						}
```
tpantelis: Can 2 zones have the same name? If not you can just use an array. Either way you can just do `Add` w/o `Contains`.

```suggestion
						zonesWithSubmarinerGW.Contains(zone.Name)
```

pkg/gcp/ocpgwdeployer.go
```diff
@@ -0,0 +1,303 @@
...
+		return nil
+	}
+
+	if len(currentGWInstanceList) < input.Gateways {
+		gatewayNodesToDeploy := input.Gateways - len(currentGWInstanceList)
+
+		reporter.Started(messageDeployGatewayNode)
+
+		for _, zone := range zones.Items {
+			if zonesWithSubmarinerGW.Contains(zone.Name) {
```
tpantelis: As mentioned in previous comments, we're still not using the same selection criteria here as above when compiling the list of existing GW zones.  So we could end up deploying on a zone whose region doesn't match `d.gcp.region` as per L84-85 above or doesn't have a public IP as per L101. I assume this would not be desired. Hence I suggest we build an `eligibleZones` list in parallel with `zonesWithSubmarinerGW`, ie if the zone has the `submarinerGatewayNodeTag` above (circa L102), gthen  add it to  `eligibleZones` and loop thru that here.

pkg/gcp/ocpgwdeployer.go
```diff
@@ -0,0 +1,303 @@
...
+		return reportFailure(reporter, err, "failed to list the zones in the project %q", d.gcp.projectID)
+	}
+
+	reporter.Succeeded(messageRetrievedZones)
+	reporter.Started(messageVerifyCurrentGWCount)
+
+	for _, zone := range zones.Items {
+		region := zone.Region[strings.LastIndex(zone.Region, "/")+1:]
+		if region != d.gcp.region {
+			continue
```
tpantelis: We really don't need to check the region here since we'll only initiate the delete below if the `submarinerGatewayNodeTag` is present. Also we don't check for the public IP here as above on L102. I think it's simpler if keep the selection criteria in one place. This code need only check for the `submarinerGatewayNodeTag`.

pkg/gcp/ocpgwdeployer.go
```diff
@@ -0,0 +1,303 @@
...
+
+			gatewayNodesToDeploy--
+			if gatewayNodesToDeploy <= 0 {
+				reporter.Succeeded(messageDeployedGatewayNode)
+				return nil
+			}
+		}
+
+		if gatewayNodesToDeploy > 0 {
+			reporter.Succeeded(messageInsufficientZonesForDeploy)
```
tpantelis: maybe a nit but should this be considered `Failed` from the reporter's standpoint?

tpantelis (changes requested):

pkg/gcp/ocpgwdeployer.go
sridhargaddam: No, two zones cannot have the same name. But a zone can have multiple instances in it.

pkg/gcp/ocpgwdeployer.go
sridhargaddam: I'm sure you meant zonesWithSubmarinerGW.Size() :)

pkg/gcp/ocpgwdeployer.go
sridhargaddam: Basically, when we list the zones in a project, it will list all the available zones in all the regions. For example - https://paste.centos.org/view/e26eead1

The validation in line 250 helps us to avoid querying the instances in other regions. Please note that with the GCP Credentials we have, we can have multiple OCP clusters in different regions (each one is uniquely identified using the InfraID). For example, you can see that when I list the instances using my GCP Creds, I can see instances that belong to other users too - https://paste.centos.org/view/e6f41036

I'm modifying the code slightly to check if the instanceName contains the infraID

pkg/gcp/ocpgwdeployer.go
sridhargaddam: Good point @tpantelis. I'm updating the code and also removing the additional check for publicIP, its sufficient if we just look for the `submarinerGatewayNodeTag` tag.

sridhargaddam (commented):
Addressed the review comments @tpantelis

pkg/gcp/ocpgwdeployer.go
tpantelis: OK so the redundant check here is for performance - makes sense. Perhaps we could put this upfront checking in an `ignoreZone` function.

pkg/gcp/ocpgwdeployer.go
tpantelis: yes 😄

pkg/gcp/ocpgwdeployer.go
```diff
@@ -257,14 +242,11 @@ func (d *ocpGatewayDeployer) Cleanup(reporter api.Reporter) error {
...
-			for _, tag := range instance.Tags.Items {
-				if tag == submarinerGatewayNodeTag {
-					err := d.deleteGateway(zone.Name)
-					if err != nil {
-						return reportFailure(reporter, err, "failed to delete gateway instance %q", instance.Name)
-					}
-
-					break
+			prefix := d.gcp.infraID + "-submariner-gw-" + zone.Name
+			if strings.HasPrefix(instance.Name, prefix) {
```
tpantelis, line 253: So now this differs with the checking circa L95. I'm not clear on why this now checks the `instance.Name` but L95 doesn't and this no longer checks `submarinerGatewayNodeTag`. It seems we could now deploy a GW on an instance w/o the desired prefix and it won't get deleted. It seems we want this check before L95 and keep the `submarinerGatewayNodeTag` check as it was here.

tpantelis (changes requested):

pkg/gcp/ocpgwdeployer.go
sridhargaddam: Basically we have two use-cases
1. Deploying new GW Nodes using the OCPMachineSet API (this PR basically implements it)
2. Modifying one (or more) of the existing Worker nodes as Gateway Nodes - I'll push a separate PR once this PR is merged.

The code in L95 will be common to both of them, hence I'm using the check `submarinerGatewayNodeTag`. Let me remove it and use the same check in this PR.

pkg/gcp/ocpgwdeployer.go
sridhargaddam: Done.

sridhargaddam (commented):
Done, updated @tpantelis

pkg/gcp/ocpgwdeployer.go
```diff
@@ -0,0 +1,289 @@
...
+			continue
+		}
+
+		instanceList, err := d.gcp.client.ListInstances(zone.Name)
+		if err != nil {
+			return reportFailure(reporter, err, "failed to list instances in zone %q of project %q", zone.Name, d.gcp.projectID)
+		}
+
+		for _, instance := range instanceList.Items {
+			prefix := d.gcp.infraID + "-submariner-gw-" + zone.Name
```
tpantelis, line 252: So the search criteria here matches that of `Deploy` 👍 . So you've changed it to check the instance name instead of the `submarinerGatewayNodeTag`. What was the reason for this and what is the purpose now of the `submarinerGatewayNodeTag`?

tpantelis (commented):
Looks good. One more comment/question.

pkg/gcp/ocpgwdeployer.go
sridhargaddam: In this PR we are adding support for deploying new Dedicated Nodes as Submariner Gateway Nodes. These nodes are created with OCP Machineset API and are named in the following manner (file pkg/gcp/gw-machineset.go) Line 25
`name: {{.InfraID}}-submariner-gw-{{.AZ}}`

So, I'm now using the name itself as an identity while deleting the Gateways.

The `submarinerGatewayNodeTag` is basically used to ensure that ExternalFirewallRules (like `nattPort UDP/4500`, `natDiscoveryPort UDP/4490`, `esp` and `ah`) are applied only on Gateway instances with `targetTags` that include `submarinerGatewayNodeTag` and not to every instance in the cluster. You can see that while creating a new dedicated Gateway node we are explicitly setting the Tag in Line 78 (pkg/gcp/gw-machineset.go) and also while creating the ExternalFirewallRules using the same Tag Line 40 (pkg/gcp/firewall_rules.go). Hope its clear.

tpantelis (approved):

aswinsuryan (approved):

vthapar (approved):

skitt (changes requested):
There are a lot of explanations of “why?” on various topics in response to the review comments in this PR; I think it would be extremely useful to have these preserved somewhere other than just the PR — either in comments in the code, or in the Submariner documentation (and if it is there already, sorry, I haven’t found it).

sridhargaddam:
> There are a lot of explanations of “why?” on various topics in response to the review comments in this PR; I think it would be extremely useful to have these preserved somewhere other than just the PR — either in comments in the code, or in the Submariner documentation (and if it is there already, sorry, I haven’t found it).

Done

sridhargaddam:
Note: Feel free to Squash the individual commits before merging.

skitt (approved):
Fantastic, thanks @sridhargaddam!

PR #104, opened 2021-09-08
Pass clientset instead of restConfig

sridhargaddam:

### DIFF
--- a/pkg/gcp/ocpgwdeployer.go
+++ b/pkg/gcp/ocpgwdeployer.go
@@ -31,7 +31,6 @@ import (
 	"google.golang.org/api/compute/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/serializer/yaml"
-	"k8s.io/client-go/rest"
 )
 
 type ocpGatewayDeployer struct {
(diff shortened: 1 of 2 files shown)


aswinsuryan (approved):

pkg/gcp/ocpgwdeployer.go
```diff
@@ -46,13 +46,13 @@ type ocpGatewayDeployer struct {
 // NewOcpGatewayDeployer returns a GatewayDeployer capable deploying gateways using OCP
 // If the supplied cloud is not a gcpCloud, an error is returned
 func NewOcpGatewayDeployer(cloud api.Cloud, msDeployer ocp.MachineSetDeployer, instanceType, image string,
-	dedicatedGWNode bool, k8sConfig *rest.Config) (api.GatewayDeployer, error) {
+	dedicatedGWNode bool, clientSet *kubernetes.Clientset) (api.GatewayDeployer, error) {
```
tpantelis: We should pass the Interface so unit tests can pass fake clientsets.

```suggestion
	dedicatedGWNode bool, client kubernetes.Interface) (api.GatewayDeployer, error) {
```

tpantelis (changes requested):

pkg/gcp/ocpgwdeployer.go
sridhargaddam: @tpantelis I hope we can simply pass k8s.k8sInterface so that you can mock it. Updating the PR, let me know if it makes sense.

tpantelis: yeah that's better.  However the `k8sInterface` is a wrapper so I don't think it would make sense to mock it - we still want to test it and we already have mocks/fakes for the API it's wrapping (ie k8s client sets). I see the `k8sInterface` as more of concrete helper functions, ie we really don't need to enclose it in a  interface. But it's fine either way - when we write unit tests we can simply use the concrete implementation rather than mocking it.

tpantelis: The important thing with this PR is that you've de-coupled the code from `rest.Config` which paves the way for unit tests.

tpantelis (approved):

PR #110, opened 2021-09-17
Add public AWS Client interface and implementation

tpantelis:
Copied from

https://github.com/open-cluster-management/submariner-addon/tree/main/pkg/cloud/aws/client


sridhargaddam:
The changes look good but its still in Draft. Planning to update it @tpantelis with UTs in the same commit?

tpantelis:
> The changes look good but its still in Draft. Planning to update it @tpantelis with UTs in the same commit?

I just intended to add the interface so the add-on can use it. I think UTs should bed another PR.

sridhargaddam (approved):

skitt (approved):

PR #115, opened 2021-09-28
Support for labelling gw nodes for generic clusters

sridhargaddam:
This PR includes support for labelling worker nodes as
Submariner gateway nodes on generic k8s clusters.

### DIFF
--- a/pkg/generic/gwdeployer.go
+++ b/pkg/generic/gwdeployer.go
@@ -1,0 +1,107 @@
+/*
+SPDX-License-Identifier: Apache-2.0
+
+Copyright Contributors to the Submariner project.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package generic
+
+import (
+	"fmt"
+
+	"github.com/submariner-io/cloud-prepare/pkg/api"
+	"github.com/submariner-io/cloud-prepare/pkg/k8s"
+	v1 "k8s.io/api/core/v1"
+)
+
+type gatewayDeployer struct {
+	k8sClient k8s.K8sInterface
+}
+
+// NewGatewayDeployer created a GatewayDeployer capable of deploying gateways to GCP
+func NewGatewayDeployer(k8sClient k8s.K8sInterface) (api.GatewayDeployer, error) {
+	return &gatewayDeployer{k8sClient: k8sClient}, nil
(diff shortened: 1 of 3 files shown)


pkg/generic/gwdeployer.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/submariner-io/cloud-prepare/pkg/k8s"
+	v1 "k8s.io/api/core/v1"
+)
+
+type gatewayDeployer struct {
+	k8sClient k8s.K8sInterface
+}
+
+// NewGatewayDeployer created a GatewayDeployer capable of deploying gateways to GCP
+func NewGatewayDeployer(k8sClient k8s.K8sInterface) (*gatewayDeployer, error) {
```
tpantelis:
```suggestion
func NewGatewayDeployer(k8sClient k8s.K8sInterface) (api.GatewayDeployer, error) {
```

pkg/generic/gwdeployer.go
```diff
@@ -0,0 +1,105 @@
...
+		return err
+	}
+
+	// Currently, we only support increasing the number of Gateway nodes which could be a valid use-case
+	// to convert a non-HA deployment to an HA deployment. We are not supporting decreasing the Gateway
+	// nodes (for now) as it might impact the datapath if we accidentally delete the active GW node.
+	if len(gwNodes.Items) < input.Gateways {
+		gatewayNodesToDeploy := input.Gateways - len(gwNodes.Items)
+
+		nonGWNodes, err := g.k8sClient.ListNodesWithLabel("!submariner.io/gateway")
```
tpantelis: The add-on also searches for the `node-role.kubernetes.io/worker` label. I assume b/c you check for master node below that this isn't necessary?

pkg/generic/gwdeployer.go
```diff
@@ -0,0 +1,105 @@
...
+	gwNodes, err := g.k8sClient.ListGatewayNodes()
+	if err != nil {
+		reporter.Failed(err)
+		return err
+	}
+
+	// Currently, we only support increasing the number of Gateway nodes which could be a valid use-case
+	// to convert a non-HA deployment to an HA deployment. We are not supporting decreasing the Gateway
+	// nodes (for now) as it might impact the datapath if we accidentally delete the active GW node.
+	if len(gwNodes.Items) < input.Gateways {
```
tpantelis:
```suggestion
	if len(gwNodes.Items) >= input.Gateways {
	    return nil
	}
```

tpantelis (changes requested):

pkg/generic/gwdeployer.go
sridhargaddam: Yes. Also JFYI, in some k8s clusters like KIND, the worker nodes are not explicitly labelled with `node-role.kubernetes.io/worker`. Hence I'm using this approach of looking for master nodes with taints that would be generic.

sridhargaddam (commented):
Thanks for reviewing the PR @tpantelis 

As the changes were minor, I did not push a separate commit but amended it to the current one.

tpantelis (approved):

aswinsuryan (approved):

PR #121, opened 2021-10-05
Add K8sInterface unit tests

tpantelis:
Also other refactoring. See individual commits for details.

### DIFF
--- a/pkg/k8s/k8siface_test.go
+++ b/pkg/k8s/k8siface_test.go
@@ -1,0 +1,275 @@
+/*
+SPDX-License-Identifier: Apache-2.0
+
+Copyright Contributors to the Submariner project.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package k8s_test
+
+import (
+	"context"
+	"sort"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/pkg/errors"
+	"github.com/submariner-io/admiral/pkg/fake"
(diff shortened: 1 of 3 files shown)


pkg/k8s/k8siface_test.go
```diff
@@ -0,0 +1,276 @@
...
+
+func testRemoveGWLabelFromWorkerNodes() {
+	t := newInterfaceTestDriver()
+
+	BeforeEach(func() {
+		t.nodes = []*corev1.Node{
+			newNode("node-1", map[string]string{k8s.SubmarinerGatewayLabel: "true"}),
+			newNode("node-2", map[string]string{k8s.SubmarinerGatewayLabel: "false"}),
+			newNode("node-3", map[string]string{k8s.SubmarinerGatewayLabel: "true", "foo": "bar"}),
+			newNode("node-4", nil),
```
sridhargaddam: nit: node-4 is unused. Its not an issue, but in case you added it for testing something, just wanted to share this observation.

pkg/k8s/k8siface_test.go
```diff
@@ -0,0 +1,276 @@
...
+			t.assertLabel(t.nodes[0].Name, k8s.SubmarinerGatewayLabel, "true")
+		})
+	})
+
+	When("the node doesn't exist", func() {
+		BeforeEach(func() {
+			t.nodes = nil
+		})
+
+		It("should not return an error", func() {
```
sridhargaddam, line 130: If the node does not exist, the API should ideally return error since Get method would fail isn't it?

sridhargaddam (commented):
Thanks for the PR @tpantelis. Looks good, just a few comments.

pkg/k8s/k8siface_test.go
tpantelis: The intent was to have a case with nil labels but the k8s list API would never return it b/c we search for the presence of the GW label. I'll remove it.

pkg/k8s/k8siface_test.go
tpantelis: The admiral `Update` function does not return an error if not found so the behavior elsewhere is to ignore it. I think that makes sense - if the caller wants to update an object then that should mean they know the object exists so a not found error means it was just deleted. We don't want callers to continuously report that and retry.

sridhargaddam (approved):

dfarrell07 (approved):

skitt (approved):

PR #123, opened 2021-10-07
Decouple gcpCloud from the GatewayDeployers

tpantelis:
The GCP `GatewayDeployers` take an `api.Cloud` instance which the code expects to be of type `gcpCloud`. It's cleaner and easier for unit tests to avoid this coupling and having to cast. Introduce a `CloudInfo` struct that contains the common fields and methods and pass that to each component.

Depends on https://github.com/submariner-io/releases/issues/241

### DIFF
--- a/pkg/gcp/ocpgwdeployer.go
+++ b/pkg/gcp/ocpgwdeployer.go
@@ -34,7 +34,8 @@ import (
 )
 
 type ocpGatewayDeployer struct {
-	gcp             *gcpCloud
+	CloudInfo
+	gcp             gcpCloud
 	msDeployer      ocp.MachineSetDeployer
 	instanceType    string
 	image           string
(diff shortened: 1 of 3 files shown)


pkg/gcp/ocpgwdeployer.go
```diff
@@ -43,28 +44,22 @@ type ocpGatewayDeployer struct {
 }
 
 // NewOcpGatewayDeployer returns a GatewayDeployer capable deploying gateways using OCP
```
sridhargaddam:
```suggestion
// NewOcpGatewayDeployer returns a GatewayDeployer capable of deploying gateways using OCP
```
minor nit

sridhargaddam (approved):

skitt (approved):

PR #127, opened 2021-10-11
Add GCP cloud unit tests

tpantelis:


pkg/gcp/gcp_cloud_test.go
```diff
@@ -0,0 +1,206 @@
...
+				t.gcpClient.EXPECT().InsertFirewallRule(projectID, gomock.Any()).DoAndReturn(func(_ string, rule *compute.Firewall) error {
+					actualRule = rule
+					return nil
+				})
+			})
+
+			It("should correctly insert it", func() {
+				Expect(retError).To(Succeed())
+
+				if actualRule != nil {
```
mkolesnik: Why would it be nil? The test should fail if this is nil.

pkg/gcp/gcp_cloud_test.go
```diff
@@ -0,0 +1,206 @@
...
+			})
+		})
+
+		Context("", func() {
+			BeforeEach(func() {
+				t.gcpClient.EXPECT().UpdateFirewallRule(projectID, ingressRuleName, gomock.Any()).Return(nil)
+			})
+
+			It("should update it", func() {
+				Expect(retError).To(Succeed())
```
mkolesnik, line 113: I wonder if you shouldn't check the sent rule as you did with insertion.

pkg/gcp/gcp_cloud_test.go
```diff
@@ -0,0 +1,206 @@
...
+		})
+	})
+
+	When("retrieval of the firewall rule fails", func() {
+		BeforeEach(func() {
+			t.gcpClient.EXPECT().GetFirewallRule(projectID, ingressRuleName).Return(nil, errors.New("fake get error"))
+		})
+
+		It("should return an error", func() {
+			Expect(retError).ToNot(Succeed())
```
mkolesnik, line 137: I wonder if it makes sense at this point to check update/insert were not called

mkolesnik (changes requested):

pkg/gcp/gcp_cloud_test.go
tpantelis: This is set in the expected call to `InsertFirewallRule` on L70.  However if it's not actually called then it will fall through and let `gomock` fail it with their nice detailed message. I could check it here but it would be redundant.

pkg/gcp/gcp_cloud_test.go
tpantelis: yeah I thought about that originally. I guess we should check it here for completeness.

pkg/gcp/gcp_cloud_test.go
tpantelis: `gomock` does that automatically, ie it fails all calls with no expectation set up.

skitt (approved):

mkolesnik (approved):

PR #146, opened 2021-11-15
Switch to Go AWS SDK v2

skitt:
See https://aws.github.io/aws-sdk-go-v2/docs/migrating/ for the
migration guide. Version 2 of the SDK features better modularisation,
more Go-idiomatic types and potentially more Go-idiomatic error
handling (albeit not on any of the errors we care about, as far as I
can determine).


pkg/aws/client/client.go
```diff
@@ -18,108 +18,124 @@ limitations under the License.
...
 	return &awsClient{
-		ec2Client: ec2.New(sess),
+		ec2Client: *ec2.NewFromConfig(cfg),
 	}, nil
 }
 
-func awsChinaEndpointResolver(service, region string, optFns ...func(*endpoints.Options)) (endpoints.ResolvedEndpoint, error) {
-	if service != route53.EndpointsID || region != "cn-northwest-1" {
-		return endpoints.DefaultResolver().EndpointFor(service, region, optFns...)
+func awsChinaEndpointResolver(service, region string) (aws.Endpoint, error) {
```
tpantelis: nit but this no longer resolves.

```suggestion
func newChinaEndpoint(service, region string) (aws.Endpoint, error) {
```

skitt: I hesitated to make it anonymous initially, I’ve just done that instead.

skitt:
Tested with https://github.com/submariner-io/submariner-operator/pull/1664.

tpantelis (approved):

dfarrell07 (approved):

PR #198, opened 2022-02-02
RHOS dedicated g/w node support

### LINKED ISSUE #167: Enhance cloud prepare to support dedicated g/w node in RHOS
**What would you like to be added**:
Enhance cloud prepare to support dedicated g/w node in RHOS
**Why is this needed**:
This needed to support Submariner in RHOS

aswinsuryan:
Signed-off-by: Aswin Suryanarayanan <aswinsuryan@gmail.com>

### DIFF
--- a/pkg/gcp/ocpgwdeployer.go
+++ b/pkg/gcp/ocpgwdeployer.go
@@ -251,7 +251,9 @@ func (d *ocpGatewayDeployer) deployGateway(zone string) error {
 	}
 
 	if d.image == "" {
-		d.image, err = d.msDeployer.GetWorkerNodeImage(machineSet, d.InfraID)
+		workerNodeList := []string{d.InfraID + "-worker-b", d.InfraID + "-worker-c", d.InfraID + "-worker-d"}
+
+		d.image, err = d.msDeployer.GetWorkerNodeImage(workerNodeList, machineSet, d.InfraID)
 		if err != nil {
 			return errors.Wrap(err, "error retrieving worker node image")
 		}
--- a/pkg/gcp/ocpgwdeployer.go
+++ b/pkg/gcp/ocpgwdeployer.go
@@ -251,6 +251,7 @@ func (d *ocpGatewayDeployer) deployGateway(zone string) error {
 	}
 
 	if d.image == "" {
+		// TODO: use machineSetClient.List() instead of hard coding.
 		workerNodeList := []string{d.InfraID + "-worker-b", d.InfraID + "-worker-c", d.InfraID + "-worker-d"}
 
 		d.image, err = d.msDeployer.GetWorkerNodeImage(workerNodeList, machineSet, d.InfraID)
--- a/pkg/k8s/k8siface.go
+++ b/pkg/k8s/k8siface.go
@@ -72,13 +72,12 @@ func (k *k8sIface) ListGatewayNodes() (*v1.NodeList, error) {
 }
 
 func (k *k8sIface) updateLabel(nodeName string, mutate func(existing *v1.Node)) error {
+	// nolint:wrapcheck // Let the caller wrap these errors.
 	client := &resource.InterfaceFuncs{
 		GetFunc: func(ctx context.Context, name string, options metav1.GetOptions) (runtime.Object, error) {
-			// nolint:wrapcheck // Let the caller wrap these errors.
 			return k.clientSet.CoreV1().Nodes().Get(ctx, name, options)
 		},
 		UpdateFunc: func(ctx context.Context, obj runtime.Object, options metav1.UpdateOptions) (runtime.Object, error) {
-			// nolint:wrapcheck // Let the caller wrap these errors.
 			return k.clientSet.CoreV1().Nodes().Update(ctx, obj.(*v1.Node), options)
 		},
 	}
--- a/pkg/k8s/k8siface.go
+++ b/pkg/k8s/k8siface.go
@@ -20,6 +20,7 @@ package k8s
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/pkg/errors"
 	"github.com/submariner-io/admiral/pkg/resource"
@@ -39,6 +40,7 @@ type Interface interface {
 	ListGatewayNodes() (*v1.NodeList, error)
 	AddGWLabelOnNode(nodeName string) error
 	RemoveGWLabelFromWorkerNodes() error
+	RemoveGWLabelFromWorkerNode(node *v1.Node) error
 }
 
 type k8sIface struct {
@@ -52,7 +54,7 @@ func NewInterface(clientSet kubernetes.Interface) Interface {
 func (k *k8sIface) ListNodesWithLabel(labelSelector string) (*v1.NodeList, error) {
 	nodes, err := k.clientSet.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{LabelSelector: labelSelector})
 	if err != nil {
-		return nil, errors.Wrap(err, "unable to list the nodes in the cluster")
+		return nil, fmt.Errorf("unable to list the nodes in the cluster, err: %w", err)
 	}
 
 	return nodes, nil
@@ -63,31 +65,32 @@ func (k *k8sIface) ListGatewayNodes() (*v1.NodeList, error) {
 
 	nodes, err := k.clientSet.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{LabelSelector: labelSelector})
 	if err != nil {
-		return nil, errors.Wrap(err, "unable to list the Gateway nodes in the cluster")
+		return nil, fmt.Errorf("unable to list the Gateway nodes in the cluster, err: %w", err)
 	}
 
 	return nodes, nil
 }
 
 func (k *k8sIface) updateLabel(nodeName string, mutate func(existing *v1.Node)) error {
-	// nolint:wrapcheck // Let the caller wrap these errors.
 	client := &resource.InterfaceFuncs{
 		GetFunc: func(ctx context.Context, name string, options metav1.GetOptions) (runtime.Object, error) {
-			return k.clientSet.CoreV1().Nodes().Get(ctx, name, options)
+			node, err := k.clientSet.CoreV1().Nodes().Get(ctx, name, options)
+			return node, errors.WithMessagef(err, "error getting node %q", name)
 		},
 		UpdateFunc: func(ctx context.Context, obj runtime.Object, options metav1.UpdateOptions) (runtime.Object, error) {
-			return k.clientSet.CoreV1().Nodes().Update(ctx, obj.(*v1.Node), options)
+			node, err := k.clientSet.CoreV1().Nodes().Update(ctx, obj.(*v1.Node), options)
+			return node, errors.WithMessagef(err, "error updating node %q ", obj.(*v1.Node).Name)
 		},
 	}
 
-	return errors.Wrap(util.Update(context.TODO(), client, &v1.Node{
+	return errors.WithMessagef(util.Update(context.TODO(), client, &v1.Node{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: nodeName,
 		},
 	}, func(existing runtime.Object) (runtime.Object, error) {
 		mutate(existing.(*v1.Node))
 		return existing, nil
-	}), "error updating node")
+	}), "failed to update node %q", nodeName)
 }
 
 func (k *k8sIface) AddGWLabelOnNode(nodeName string) error {
@@ -105,19 +108,22 @@ func (k *k8sIface) AddGWLabelOnNode(nodeName string) error {
 func (k *k8sIface) RemoveGWLabelFromWorkerNodes() error {
 	gwNodeList, err := k.clientSet.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{LabelSelector: SubmarinerGatewayLabel})
 	if err != nil {
-		return errors.Wrap(err, "error listing gateway nodes")
+		return errors.WithMessage(err, "error listing submariner gateway nodes")
 	}
 
-	for i := range gwNodeList.Items {
-		node := &gwNodeList.Items[i]
-		err = k.updateLabel(node.Name, func(existing *v1.Node) {
-			delete(existing.Labels, SubmarinerGatewayLabel)
-		})
-
+	gwNodes := gwNodeList.Items
+	for i := range gwNodes {
+		err = k.RemoveGWLabelFromWorkerNode(&gwNodes[i])
 		if err != nil {
-			return err
+			return errors.WithMessagef(err, "error removing the label from the gateway node %q", gwNodes[i].Name)
 		}
 	}
 
 	return nil
+}
+
+func (k *k8sIface) RemoveGWLabelFromWorkerNode(node *v1.Node) error {
+	return k.updateLabel(node.Name, func(existing *v1.Node) {
+		delete(existing.Labels, SubmarinerGatewayLabel)
+	})
 }
--- a/pkg/k8s/k8siface.go
+++ b/pkg/k8s/k8siface.go
@@ -74,12 +74,12 @@ func (k *k8sIface) ListGatewayNodes() (*v1.NodeList, error) {
 func (k *k8sIface) updateLabel(nodeName string, mutate func(existing *v1.Node)) error {
 	client := &resource.InterfaceFuncs{
 		GetFunc: func(ctx context.Context, name string, options metav1.GetOptions) (runtime.Object, error) {
-			node, err := k.clientSet.CoreV1().Nodes().Get(ctx, name, options)
-			return node, errors.Wrap(err, fmt.Sprintf("error getting node %q", name))
+			// nolint:wrapcheck // Let the caller wrap these errors.
+			return k.clientSet.CoreV1().Nodes().Get(ctx, name, options)
 		},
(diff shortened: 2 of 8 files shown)


pkg/k8s/k8siface.go
```diff
@@ -52,7 +54,7 @@ func NewInterface(clientSet kubernetes.Interface) Interface {
 func (k *k8sIface) ListNodesWithLabel(labelSelector string) (*v1.NodeList, error) {
 	nodes, err := k.clientSet.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{LabelSelector: labelSelector})
 	if err != nil {
-		return nil, errors.Wrap(err, "unable to list the nodes in the cluster")
+		return nil, fmt.Errorf("unable to list the nodes in the cluster, err: %w", err)
```
tpantelis: Why was this changed?

pkg/k8s/k8siface.go
```diff
@@ -63,31 +65,32 @@ func (k *k8sIface) ListGatewayNodes() (*v1.NodeList, error) {
 
 	nodes, err := k.clientSet.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{LabelSelector: labelSelector})
 	if err != nil {
-		return nil, errors.Wrap(err, "unable to list the Gateway nodes in the cluster")
+		return nil, fmt.Errorf("unable to list the Gateway nodes in the cluster, err: %w", err)
```
tpantelis: ...and this?

pkg/k8s/k8siface.go
```diff
@@ -63,31 +65,32 @@ func (k *k8sIface) ListGatewayNodes() (*v1.NodeList, error) {
...
 	client := &resource.InterfaceFuncs{
 		GetFunc: func(ctx context.Context, name string, options metav1.GetOptions) (runtime.Object, error) {
-			return k.clientSet.CoreV1().Nodes().Get(ctx, name, options)
+			node, err := k.clientSet.CoreV1().Nodes().Get(ctx, name, options)
+			return node, errors.WithMessagef(err, "error getting node %q", name)
 		},
 		UpdateFunc: func(ctx context.Context, obj runtime.Object, options metav1.UpdateOptions) (runtime.Object, error) {
-			return k.clientSet.CoreV1().Nodes().Update(ctx, obj.(*v1.Node), options)
+			node, err := k.clientSet.CoreV1().Nodes().Update(ctx, obj.(*v1.Node), options)
+			return node, errors.WithMessagef(err, "error updating node %q ", obj.(*v1.Node).Name)
```
tpantelis: As per the prior comment, I don't think we need to wrap here as these functions are pass-throughs.

pkg/k8s/k8siface.go
```diff
@@ -63,31 +65,32 @@ func (k *k8sIface) ListGatewayNodes() (*v1.NodeList, error) {
...
 		},
 	}
 
-	return errors.Wrap(util.Update(context.TODO(), client, &v1.Node{
+	return errors.WithMessagef(util.Update(context.TODO(), client, &v1.Node{
```
tpantelis: Why was this changed from `Wrap` to `WithMessage`? Same in other places?

pkg/rhos/ocpgwdeployer.go
```diff
@@ -18,28 +18,126 @@ limitations under the License.
...
+
+	unstructDecoder := yaml.NewDecodingSerializer(unstructured.UnstructuredJSONScheme)
+	machineSet := &unstructured.Unstructured{}
+	_, _, err = unstructDecoder.Decode(gatewayYAML, nil, machineSet)
+
+	if err != nil {
+		return nil, errors.WithMessage(err, "error decoding message gateway yaml")
+	}
+
+	return machineSet, nil
```
tpantelis:
```suggestion
	return machineSet, errors.WithMessage(err, "error decoding message gateway yaml")
```

tpantelis (changes requested):

pkg/k8s/k8siface.go
```diff
@@ -70,13 +72,14 @@ func (k *k8sIface) ListGatewayNodes() (*v1.NodeList, error) {
...
 	client := &resource.InterfaceFuncs{
 		GetFunc: func(ctx context.Context, name string, options metav1.GetOptions) (runtime.Object, error) {
-			return k.clientSet.CoreV1().Nodes().Get(ctx, name, options)
+			node, err := k.clientSet.CoreV1().Nodes().Get(ctx, name, options)
+			return node, errors.Wrap(err, fmt.Sprintf("error getting node %q", name))
 		},
 		UpdateFunc: func(ctx context.Context, obj runtime.Object, options metav1.UpdateOptions) (runtime.Object, error) {
-			return k.clientSet.CoreV1().Nodes().Update(ctx, obj.(*v1.Node), options)
+			node, err := k.clientSet.CoreV1().Nodes().Update(ctx, obj.(*v1.Node), options)
+			return node, errors.Wrap(err, fmt.Sprintf("error updating node %q ", obj.(*v1.Node).Name))
```
tpantelis: As mentioned before, these are pass-throughs so we don't need to wrap - let the caller do it.

pkg/k8s/k8siface.go
```diff
@@ -105,19 +108,22 @@ func (k *k8sIface) AddGWLabelOnNode(nodeName string) error {
...
-		err = k.updateLabel(node.Name, func(existing *v1.Node) {
-			delete(existing.Labels, SubmarinerGatewayLabel)
-		})
-
+	gwNodes := gwNodeList.Items
+	for i := range gwNodes {
+		err = k.RemoveGWLabelFromWorkerNode(&gwNodes[i])
 		if err != nil {
-			return err
+			return errors.Wrap(err, fmt.Sprintf("error removing the label from the gateway node %q", gwNodes[i].Name))
```
tpantelis, line 117:
```suggestion
			return errors.Wrapf(err, "error removing the label from the gateway node %q", gwNodes[i].Name)
```

pkg/ocp/machinesets.go
```diff
@@ -89,17 +88,24 @@ func (msd *k8sMachineSetDeployer) GetWorkerNodeImage(machineSet *unstructured.Un
...
+				image, _, _ := unstructured.NestedString(disk, "image")
+				if image != "" {
+					return image, nil
+				}
+			}
 		}
 	}
 
-	return "", fmt.Errorf("could not find the image of one of the worker nodes on GCP infra %q", infraID)
+	return "", errors.Wrap(err, fmt.Sprintf("could not retrieve the image of one of the worker nodes on gcp infra %q", infraID))
```
tpantelis: There's no error in scope that would be non-nil here - `fmt.Errorf` is appropriate here..

pkg/rhos/ocpgwdeployer.go
```diff
@@ -65,55 +159,73 @@ func (d *ocpGatewayDeployer) Deploy(input api.GatewayDeployInput, reporter api.R
...
+				alreadyTagged := nodes[i].GetLabels()[submarinerGatewayNodeTag]
+				if alreadyTagged == "true" {
+					continue
+				}
+
+				reporter.Started(fmt.Sprintf("Configuring worker node %q as Submariner gateway node", nodes[i].Name))
+
+				err := d.K8sClient.AddGWLabelOnNode(nodes[i].Name)
+				if err != nil {
+					return errors.Wrap(err, fmt.Sprintf("failed to label the node %q as Submariner gateway node", nodes[i].Name))
```
tpantelis:
```suggestion
					return errors.Wrapf(err, "failed to label the node %q as Submariner gateway node", nodes[i].Name)
```

pkg/rhos/ocpgwdeployer.go
```diff
@@ -125,28 +237,51 @@ func (d *ocpGatewayDeployer) Cleanup(reporter api.Reporter) error {
 
 	computeClient, err := openstack.NewComputeV2(d.Client, gophercloud.EndpointOpts{Region: d.Region})
 	if err != nil {
-		return errors.WithMessagef(err, "error creating the compute client for the region: %q", d.Region)
+		return errors.Wrap(err, fmt.Sprintf("error creating the compute client for the region: %q", d.Region))
```
tpantelis:
```suggestion
		return errors.Wrapf(err, "error creating the compute client for the region: %q", d.Region)
```

pkg/rhos/ocpgwdeployer.go
```diff
@@ -125,28 +237,51 @@ func (d *ocpGatewayDeployer) Cleanup(reporter api.Reporter) error {
...
-		return errors.WithMessage(err, "failed to remove labels from worker node")
+		// If the instance name matches with d.InfraID + "-submariner-gw-", it implies that
+		// the gateway node was deployed using the OCPMachineSet API otherwise it's an existing worker node.
+		prefix := d.InfraID + "-submariner-gw-"
+		if strings.HasPrefix(gwNodes[i].Name, prefix) {
+			reporter.Started(fmt.Sprintf("Deleting the gateway instance %q", gwNodes[i].Name))
+
+			err = d.deleteGateway(strconv.Itoa(i))
+			if err != nil {
+				return errors.Wrap(err, fmt.Sprintf("error deleting the Submariner gateway security group rules from node: %q",
```
tpantelis:
```suggestion
				return errors.Wrapf(err, "error deleting the Submariner gateway security group rules from node: %q",
```

pkg/rhos/ocpgwdeployer.go
```diff
@@ -125,28 +237,51 @@ func (d *ocpGatewayDeployer) Cleanup(reporter api.Reporter) error {
...
+					gwNodes[i].Name))
+			}
+
+			reporter.Succeeded("Successfully deleted the instance")
+		} else {
+			err = d.removeGWFirewallRules(groupName, gwNodes[i].Name, computeClient)
+			if err != nil {
+				return errors.Wrap(err, fmt.Sprintf("error deleting the Submariner gateway security group rules from node: %q",
```
tpantelis:
```suggestion
				return errors.Wrapf(err, "error deleting the Submariner gateway security group rules from node: %q",
```

tpantelis (changes requested):

pkg/k8s/k8siface.go
aswinsuryan: Linter was throwing an error if not wrapped

pkg/k8s/k8siface.go:81:11: error returned from interface method should be wrapped: sig: func (k8s.io/client-go/kubernetes/typed/core/v1.NodeInterface).Update(ctx context.Context, node *k8s.io/api/core/v1.Node, opts k8s.io/apimachinery/pkg/apis/meta/v1.UpdateOptions) (*k8s.io/api/core/v1.Node, error) (wrapcheck)
			return k.clientSet.CoreV1().Nodes().Update(ctx, obj.(*v1.Node), options)

tpantelis: Right - that's why I previously had "_// nolint:wrapcheck // Let the caller wrap these errors._" on L73.

aswinsuryan: oh ok i got it now, added the comment .

pkg/k8s/k8siface.go
```diff
@@ -70,12 +72,13 @@ func (k *k8sIface) ListGatewayNodes() (*v1.NodeList, error) {
...
 
 func (k *k8sIface) updateLabel(nodeName string, mutate func(existing *v1.Node)) error {
-	// nolint:wrapcheck // Let the caller wrap these errors.
 	client := &resource.InterfaceFuncs{
 		GetFunc: func(ctx context.Context, name string, options metav1.GetOptions) (runtime.Object, error) {
+			// nolint:wrapcheck // Let the caller wrap these errors.
 			return k.clientSet.CoreV1().Nodes().Get(ctx, name, options)
 		},
 		UpdateFunc: func(ctx context.Context, obj runtime.Object, options metav1.UpdateOptions) (runtime.Object, error) {
+			// nolint:wrapcheck // Let the caller wrap these errors.
```
tpantelis: This works but I meant to put the comment back before L75 where it was previously.

aswinsuryan: I corrected it.

tpantelis (approved):

pkg/gcp/ocpgwdeployer.go
```diff
@@ -251,7 +251,9 @@ func (d *ocpGatewayDeployer) deployGateway(zone string) error {
 	}
 
 	if d.image == "" {
-		d.image, err = d.msDeployer.GetWorkerNodeImage(machineSet, d.InfraID)
+		workerNodeList := []string{d.InfraID + "-worker-b", d.InfraID + "-worker-c", d.InfraID + "-worker-d"}
```
sridhargaddam, line 255: Can we safely assume that nodes will be labelled in this manner (i.e., `-worker-x`)?
Also, here we are hard-coding the list to just four nodes. Is there a better way to simply query the worker nodes based on the labels and pick one of them?

pkg/ocp/machinesets.go
```diff
@@ -69,15 +69,14 @@ func (msd *k8sMachineSetDeployer) clientFor(obj runtime.Object) (dynamic.Resourc
...
-func (msd *k8sMachineSetDeployer) GetWorkerNodeImage(machineSet *unstructured.Unstructured, infraID string) (string, error) {
+func (msd *k8sMachineSetDeployer) GetWorkerNodeImage(workerNodeList []string, machineSet *unstructured.Unstructured,
+	infraID string) (string, error) {
 	machineSetClient, err := msd.clientFor(machineSet)
 	if err != nil {
 		return "", err
 	}
 
 	// TODO: After implementing a ListAll method in admiral, modify this code accordingly.
-	workerNodeList := []string{infraID + "-worker-b", infraID + "-worker-c", infraID + "-worker-d"}
```
sridhargaddam, line 79: Can we remove the above TODO comment as it does not seem to be relevant any more?

pkg/ocp/machinesets.go
```diff
@@ -89,17 +88,24 @@ func (msd *k8sMachineSetDeployer) GetWorkerNodeImage(machineSet *unstructured.Un
...
+				image, _, _ := unstructured.NestedString(disk, "image")
+				if image != "" {
+					return image, nil
+				}
+			}
 		}
 	}
 
-	return "", fmt.Errorf("could not find the image of one of the worker nodes on GCP infra %q", infraID)
+	return "", fmt.Errorf("could not retrieve the image of one of the worker nodes on rhos infra %q", infraID)
```
sridhargaddam:
```suggestion
	return "", fmt.Errorf("could not retrieve the image of one of the worker nodes from the infra %q", infraID)
```

As this is used even in GCP, lets try to use some generic message.

sridhargaddam (changes requested):

pkg/ocp/machinesets.go
aswinsuryan: Done.

pkg/ocp/machinesets.go
aswinsuryan: Moved the TODO to the respective callers.

pkg/gcp/ocpgwdeployer.go
aswinsuryan: I was not sure why it was done this way , but as per the TODO it is till admiral adds an API.

tpantelis: I'm not sure I follow. Why what was done? Lines 254-255 are new, added by this PR. Can you be more specific re: a `ListAll` method in admiral, ie specifically where would it be added and what would it do?

aswinsuryan: These are moved from the pkg/ocp/machinesets.go to this file and now the workerNodeList is passed as argument  to msDeployer.GetWorkerNodeImage so that the method can be reused. This may be referring to a list all function to see all the worker nodes in the infra.

tpantelis: OK I still don't see where such a `ListAll` function would go or why it would be needed in admiral. Not trying to nit but unless we understand what the TODO means, I'd rather not propagate it.

aswinsuryan: I removed the TODO.

sridhargaddam: Basically, we are using the admiral dynamic Interface API to Get the machineSet details associated with the worker-node. Currently, the admiral API does not have the listAll method (https://github.com/submariner-io/admiral/blob/69a40598bdd63f570d0e695b2fd1a3ebdeb8f0f0/pkg/resource/interface.go#L28), hence we are using this approach to get the machineSet details of the associated worker nodes. If we have a listAll method in admiral, we can remove this hard-coding of the workerNodeList.

@aswinsuryan thanks for updating the PR. It looks good. Can you please add a TODO with the above notes so that once Admiral is enhanced we can update this code accordingly?

tpantelis: I'm confused. I don't see the admiral `Resource` interface being used anywhere in this code path. This code came from `k8sMachineSetDeployer.GetWorkerNodeImage` which uses a `dynamic.Interface` obtained from

    `machineSetClient, err := msd.clientFor(machineSet)`

 so the standard K8s List method is available.

sridhargaddam: Yes @tpantelis. I am not able to recollect the exact reason why that TODO was added. But it looks like we can use machineSetClient.List() method to get the list of worker nodes and remove the hardcoded list.

sridhargaddam (approved):

pkg/gcp/ocpgwdeployer.go
aswinsuryan: I have added a TODO and raised this issue to track https://github.com/submariner-io/cloud-prepare/issues/201

PR #358, opened 2022-08-18
Limit Azure machine name lengths to 40 characters

skitt:
To avoid hitting the 63-character limit on "machine IP names", which
are generated using the provided machine names, limit the latter to 40
characters.

The MachineName function is exported for the unit test to limit the
changes required for this fix; a proper test would require a full
driver setup as in other packages.

### DIFF
--- a/pkg/azure/ocpgwdeployer.go
+++ b/pkg/azure/ocpgwdeployer.go
@@ -39,7 +39,7 @@ import (
 )
 
 const (
-	submarinerGatewayGW      = "subgw"
+	submarinerGatewayGW      = "-subgw-"
 	azureVirtualMachines     = "virtualMachines"
 	topologyLabel            = "topology.kubernetes.io/zone"
 	submarinerGatewayNodeTag = "submariner-io-gateway-node"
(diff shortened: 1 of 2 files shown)


pkg/azure/ocpgwdeployer.go
```diff
@@ -267,16 +267,35 @@ func (d *ocpGatewayDeployer) initMachineSet(name, zone string) (*unstructured.Un
...
 
 	return errors.Wrapf(d.msDeployer.Deploy(machineSet), "error deploying machine set %q", machineSet.GetName())
 }
 
+func MachineName(infraID, region, zone string) string {
+	// Limit the name length to 40 characters to ensure we don't hit the 63-character limit
+	// when generating the "machine public IP name".
+	// At most 6 characters for the zone, at most 12 for the region and zone combined,
+	// at most 32 for the infra id, region and zone combined.
+	// We add "-subgw-", 7 characters, for a total of 40 with the hyphen between region and zone.
```
mkolesnik: Wouldn't it be better as a function doc?

pkg/azure/ocpgwdeployer_test.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/submariner-io/cloud-prepare/pkg/azure"
+)
+
+func TestAzure(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Azure Suite")
+}
+
+var _ = Describe("Azure", func() {
+	Describe("Generating machine names", func() {
```
tpantelis:
```suggestion
var _ = Describe("OCP Gateway Deployer", func() {
	Describe("MachineName", func() {
```

pkg/azure/ocpgwdeployer.go
tpantelis: Why do we specifically limit the zone to 6 and the region to 12?

pkg/azure/ocpgwdeployer_test.go
```diff
@@ -0,0 +1,42 @@
...
+		It("should be at most 40 characters in length", func() {
+			Expect(len(azure.MachineName("acmqe-clc-auto-azure6-fd55b", "centralus", "3")) <= 40).To(BeTrue())
```
tpantelis: On failure, we can make the output clearer:

```suggestion
			Expect(len(azure.MachineName("acmqe-clc-auto-azure6-fd55b", "centralus", "3"))).To(BeNumerically("<=", 40))
```

pkg/azure/ocpgwdeployer.go
```diff
@@ -267,16 +267,35 @@ func (d *ocpGatewayDeployer) initMachineSet(name, zone string) (*unstructured.Un
...
+
+	if len(region) > 12-len(zone) {
+		region = region[0 : 12-len(zone)]
+	}
+
+	if len(infraID) > 32-len(zone)-len(region) {
+		infraID = infraID[0 : 32-len(zone)-len(region)]
+	}
+
+	return infraID + submarinerGatewayGW + region + "-" + zone
```
tpantelis, line 302: So for inputs:

infraID: "1234"
region: "123"
zode:    "1234567"

the combined length of the three would not exceed 32 but we'd still truncate the zone. Perhaps we should first check the combined length, ie don't truncate if we don't need to:

```Go
if len(zone) + len(region) + len(infraID) > 32 {
    if len(zone) > 6 {
        zone = zone[0:6]
    }
    ...
}

return infraID + submarinerGatewayGW + region + "-" + zone
```

tpantelis (changes requested):

tpantelis (approved):

aswinsuryan (approved):

dfarrell07 (approved):

PR #385, opened 2022-09-11
Add load balance support for aws

### LINKED ISSUE #1943: SM deployment (using LB) fails on OCP/AWS when number of gateways set in cloud prepare is > 0
**What happened**:
A. I deployed OCP on AWS, check [1] for version details
B. Run subctl prepare using [2] command
C. Deploy SM with LoadBalancer service using subctl (0.13) CLI 

SM deployment failed and the root cause was K8S service-controller failed to ensure the SM gateway LB service, with [3] error.

Running 'subctl prepare' with --gateways 0 solved the problem, but we need to find a way how to label more than one node with 'submariner.io/gateway=true' to enable GW HA.

[1] 
$ oc version 
Client Version: 4.8.11
Server Version: 4.10.10
Kubernetes Version: v1.23.5+9ce5071
$ 

[2] 
subctl cloud prepare aws  --ocp-metadata aws-cluster-b/metadata.json

[3] 
```
$ oc describe  svc -n submariner-operator  submariner-gateway 
Name:                     submariner-gateway
Namespace:                submariner-operator
Labels:                   <none>
Annotations:              service.beta.kubernetes.io/aws-load-balancer-type: nlb
Selector:                 app=submariner-gateway,gateway.submariner.io/status=active
Type:                     LoadBalancer
IP Family Policy:         SingleStack
IP Families:              IPv4
IP:                       172.30.25.145
IPs:                      172.30.25.145
Port:                     cable-encaps  4500/UDP
TargetPort:               4500/UDP
NodePort:                 cable-encaps  30531/UDP
Endpoints:                <none>
Port:                     natt-discovery  4490/UDP
TargetPort:               4490/UDP
NodePort:                 natt-discovery  30766/UDP
Endpoints:                <none>
Session Affinity:         None
External Traffic Policy:  Local
HealthCheck NodePort:     31484
Events:
  Type     Reason                  Age                   From                Message
  ----     ------                  ----                  ----                -------
  Normal   EnsuringLoadBalancer    2m11s (x12 over 32m)  service-controller  Ensuring load balancer
  Warning  SyncLoadBalancerFailed  2m10s (x12 over 32m)  service-controller  Error syncing load balancer: failed to ensure load balancer: Multiple tagged security groups found for instance i-0acd2ff8a57ea9b85; ensure only the k8s security group is tagged; the tagged groups were sg-00e5ee55f67f27562(terraform-20220727163351506400000002) sg-0208130199687d637(yboaron-cluster-b-mqv58-submariner-gw-sg)
$
```

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Diagnose information (use `subctl diagnose all`):
- Gather information (use `subctl gather`):
- Cloud provider or hardware configuration:
- Install tools:
- Others:

yboaron:
Update aws code to support load balance mode.

Fixes: https://github.com/submariner-io/submariner/issues/1943


pkg/aws/ocpgwdeployer.go
```diff
@@ -83,14 +86,18 @@ func (d *ocpGatewayDeployer) Deploy(input api.GatewayDeployInput, status reporte
 
 	status.Success(messageValidatedPrerequisites)
 
-	status.Start("Creating Submariner gateway security group")
+	gatewaySG := ""
```
tpantelis, line 89: This can be declared local on L94.

```suggestion
```

pkg/aws/ocpgwdeployer.go
```diff
@@ -83,14 +86,18 @@ func (d *ocpGatewayDeployer) Deploy(input api.GatewayDeployInput, status reporte
...
 
-	gatewaySG, err := d.aws.createGatewaySG(vpcID, input.PublicPorts)
-	if err != nil {
-		return status.Error(err, "unable to create gateway")
-	}
+	if !d.useLoadBalancer {
+		status.Start("Creating Submariner gateway security group")
 
-	status.Success("Created Submariner gateway security group %s", gatewaySG)
+		gatewaySG, err = d.aws.createGatewaySG(vpcID, input.PublicPorts)
```
tpantelis, line 94:
```suggestion
		gatewaySG, err := d.aws.createGatewaySG(vpcID, input.PublicPorts)
```

tpantelis (changes requested):

pkg/aws/ocpgwdeployer.go
yboaron: gatewaySG also used in L141 outside of if scope.

yboaron:
cc @aswinsuryan @sridhargaddam

pkg/aws/ocpgwdeployer.go
tpantelis: I missed that but should L141 even execute if `gatewaySG` is not set?

yboaron: >I missed that but should L141 even execute if gatewaySG is not set?

Yep, it should. deployGateway should be called regardless of LB configuration and sg.

tpantelis (approved):

pkg/aws/gw-machineset.go
```diff
@@ -65,12 +65,14 @@ spec:
                 - name: tag:Name
                   values:
                     - {{.InfraID}}-worker-sg
+          {{if not .UseLoadBalancer}}
```
mkolesnik:
```suggestion
          {{- if not .UseLoadBalancer }}
```

You need a `-` otherwise this will be rendered as an empty line, also space is necessary in this case.
https://pkg.go.dev/text/template#hdr-Text_and_spaces

You can test the template here - https://gotemplate.io/

pkg/aws/gw-machineset.go
```diff
@@ -65,12 +65,14 @@ spec:
...
                     - {{.SecurityGroup}}
           subnet:
             filters:
               - name: tag:Name
                 values:
                   - {{.PublicSubnet}}
+          {{ end }}
```
mkolesnik:
```suggestion
          {{- end }}
```

Here also you need the `-`

pkg/aws/ocpgwdeployer.go
```diff
@@ -83,14 +86,18 @@ func (d *ocpGatewayDeployer) Deploy(input api.GatewayDeployInput, status reporte
...
+		if err != nil {
+			return status.Error(err, "unable to create gateway")
+		}
+
+		status.Success("Created Submariner gateway security group %s", gatewaySG)
+	}
 
 	subnets, err := d.aws.getSubnetsSupportingInstanceType(publicSubnets, d.instanceType)
```
mkolesnik, line 102: You're not actually using the subnet in the deployed machineset (which is fine), so just tagging them for submariner but not actually connecting to them makes no sense..

In fact, you need to just deploy gateway per AZ, as many as requested, so the code path is quite different.

pkg/aws/ocpgwdeployer.go
```diff
@@ -157,6 +164,10 @@ func (d *ocpGatewayDeployer) validateDeployPrerequisites(vpcID string, input api
 		return utilerrors.NewAggregate(errs)
 	}
 
+	if d.useLoadBalancer {
+		return nil
```
mkolesnik, line 168: Actually for LB the `validateCreateSecGroup` call is not needed, so the function needs to be restructured

mkolesnik (changes requested):

pkg/aws/ocpgwdeployer.go
yboaron: That's correct. so currently (non LB )we have :
- Create flow:
  1. need to find an appropriate subnet  with some logic related to AZ 
  2. tag this subnet 
  3. Create public SG
  4. Create machineset manifest (including public subnet + public SG)

- Delete flow:
 1. Get tagged subnet 
 2. Extract AZ from tagged subnet
 3. Delete machineset 
 
IIUC, tagging a subnet is something like annotation, so I thought instead of having 2 separate flows for create and delete , maintain a single flow (as possible) with some changes, for LB:

Create:
same flow, with the following changes:
Skip step 3 
and update step 4 (create machineset without SG and public subnet)

Delete:
Same flow (retrieve the AZ from subnet)

If you think that tagging a subnet for LB case is not acceptable then we need to come up with another approach, but if it's something equivalent to annotation maybe it should be OK?

mkolesnik: Generally a tag in AWS is just metadata so it won't affect the functionality, that is true.

However, the problem is that the amount of available public subnets could be lower that the requested amount of gateways, so to properly support LB mode it's best to avoid relying on public subnets altogether, since we don't actually need them.

Unfortunately, this would necessitate a separate code path for LB, but IMHO it's better than relying on subnets, which in an LB case is an unnecessary constraint.

dfarrell07:
Rebasing to fix

```
Checking #385
  No dependencies found. Running clean-up
  Error: HttpError: Validation Failed: {"resource":"Status","code":"custom","message":"This SHA and context has reached the maximum number of statuses."}
```

https://github.com/submariner-io/cloud-prepare/actions/runs/4863187727/jobs/8670539853

tpantelis:
@yboaron Is this still relevant?

yboaron:
> @yboaron Is this still relevant?

This PR was one of the tasks under reducing cloud preparation capabilities epic, since we are not putting effort in this direction, I guess this PR can be closed.

PR #446, opened 2022-11-02
Don't tag Submariner SG as cluster-owned

skitt:
Setting the Submariner SG as cluster-owned results in multiple SGs being found for load-balancer services, which isn't supported. This was previously fixed in commit 54e25267a87e ("Remove the cluster owned tag from submariner sg") in the Submariner project.

See https://github.com/kubernetes/kubernetes/issues/73906 and https://bugzilla.redhat.com/show_bug.cgi?id=2139477


tpantelis (approved):

nyechiel:
It looks like UT are failing.

maayanf24 (approved):

mkolesnik (approved):

PR #506, opened 2022-12-07
Use k8s.io/utils/pointer instead of go-autorest/to

skitt:
https://github.com/Azure/go-autorest is deprecated and will no longer be supported after March 31, 2023; see
https://azure.microsoft.com/en-gb/updates/support-for-azure-sdk-libraries-that-do-not-conform-to-our-current-azure-sdk-guidelines-will-be-retired-as-of-31-march-2023/

We only use go-autorest to create pointers; this replaces such calls with calls to equivalent k8s.io/utils functions.


skitt:
This should be backported to all affected supported branches.

tpantelis (approved):

dfarrell07 (approved):

PR #508, opened 2022-12-09
Fix issues in cloud prepare for Azure and RHOS

aswinsuryan:
Fix issues in the cloud prepare for Azure and RHOS

1) Fix Azure machine set name now changed to region + uuid
and limited to 20 characters to prevent issues long cluster
name

2) Machine set creation and deletion logic updated. Now it lists
machines set to check for existing dedicated nodes and clean up
gateway nodes along with tagged g/w nodes. Earlier it listed
only tagged nodes. This resulted in the creation of multiple g/w nodes
cases when cloud prepare is run multiple times due to a delay in the creation
g/w nodes from the machine set.

Fixes: https://github.com/stolostron/backlog/issues/25111

### DIFF
--- a/pkg/azure/ocpgwdeployer.go
+++ b/pkg/azure/ocpgwdeployer.go
@@ -21,7 +21,6 @@ package azure
 import (
 	"bytes"
 	"context"
-	"strings"
 	"text/template"
 	"time"
 
@@ -32,16 +31,15 @@ import (
 	"github.com/submariner-io/admiral/pkg/stringset"
 	"github.com/submariner-io/cloud-prepare/pkg/api"
 	"github.com/submariner-io/cloud-prepare/pkg/ocp"
-	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/serializer/yaml"
+	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/utils/pointer"
 )
 
 const (
-	submarinerGatewayGW      = "-subgw-"
+	submarinerGatewayGW      = "subgw-"
 	azureVirtualMachines     = "virtualMachines"
-	topologyLabel            = "topology.kubernetes.io/zone"
 	submarinerGatewayNodeTag = "submariner-io-gateway-node"
 )
 
@@ -79,40 +77,36 @@ func (d *ocpGatewayDeployer) Deploy(input api.GatewayDeployInput, status reporte
 
 	status.Start("Deploying gateway node")
 
-	nsgClient, err := d.getNsgClient()
+	nsgClient, nwClient, pubIPClient, err := d.getClients(status)
 	if err != nil {
-		return status.Error(err, "Failed to get network security groups client")
-	}
-
-	nwClient, err := d.getInterfacesClient()
-	if err != nil {
-		return status.Error(err, "Failed to get network interfaces client")
-	}
-
-	pubIPClient, err := d.getPublicIPClient()
-	if err != nil {
-		return status.Error(err, "Failed to get network public IP addresses client")
+		return err
 	}
 
 	groupName := d.InfraID + externalSecurityGroupSuffix
+
+	machineSets, err := d.msDeployer.List()
+	if err != nil {
+		return status.Error(err, "error getting the gateway machinesets")
+	}
 
 	gwNodes, err := d.azure.K8sClient.ListGatewayNodes()
 	if err != nil {
 		return errors.Wrap(err, "error getting the gateway node")
 	}
 
-	// Open the g/w ports and assign public-ip if not already done for manually tagged nodes if any
 	gwNodeItems := gwNodes.Items
-	gatewayNodesToDeploy := input.Gateways - len(gwNodeItems)
+	taggedExistingNodes := ocp.RemoveDuplicates(machineSets, gwNodeItems)
+	gatewayNodesToDeploy := input.Gateways - len(machineSets) - len(taggedExistingNodes)
 
-	if len(gwNodeItems) != 0 || gatewayNodesToDeploy != 0 {
+	if len(machineSets) != 0 || gatewayNodesToDeploy != 0 {
 		if err := d.createGWSecurityGroup(groupName, input.PublicPorts, nsgClient); err != nil {
 			return status.Error(err, "creating gateway security group failed")
 		}
 	}
 
+	// Open the g/w ports and assign public-ip if not already done for manually tagged nodes if any
 	for i := range gwNodeItems {
-		if err = d.prepareGWInterface(gwNodeItems[i].Name, groupName, nsgClient, nwClient, pubIPClient); err != nil {
+		if err = d.prepareGWInterface(gwNodeItems[i].GetName(), groupName, nsgClient, nwClient, pubIPClient); err != nil {
 			return status.Error(err, "failed to open the Submariner gateway port for already existing nodes")
 		}
 	}
@@ -131,7 +125,7 @@ func (d *ocpGatewayDeployer) Deploy(input api.GatewayDeployInput, status reporte
 	}
 
 	if d.dedicatedGWNode {
-		err = d.deployDedicatedGWNode(gwNodeItems, gatewayNodesToDeploy, status)
+		err = d.deployDedicatedGWNode(machineSets, gatewayNodesToDeploy, status)
 	} else {
 		err = d.tagExistingNode(nsgClient, nwClient, pubIPClient, gatewayNodesToDeploy, status)
 	}
@@ -143,7 +137,8 @@ func (d *ocpGatewayDeployer) Deploy(input api.GatewayDeployInput, status reporte
 	return err
 }
 
-func (d *ocpGatewayDeployer) deployDedicatedGWNode(gwNodes []v1.Node, gatewayNodesToDeploy int, status reporter.Interface,
+func (d *ocpGatewayDeployer) deployDedicatedGWNode(gwNodes []unstructured.Unstructured, gatewayNodesToDeploy int,
+	status reporter.Interface,
 ) error {
 	az, err := d.getAvailabilityZones(gwNodes)
 	if err != nil || az.Size() == 0 {
@@ -267,7 +262,7 @@ func (d *ocpGatewayDeployer) initMachineSet(name, zone string) (*unstructured.Un
 }
 
 func (d *ocpGatewayDeployer) deployGateway(zone string) error {
-	machineSet, err := d.initMachineSet(MachineName(d.azure.InfraID, d.azure.Region, zone), zone)
+	machineSet, err := d.initMachineSet(MachineName(d.azure.Region), zone)
 	if err != nil {
 		return err
 	}
@@ -276,37 +271,29 @@ func (d *ocpGatewayDeployer) deployGateway(zone string) error {
 }
 
 // MachineName generates a machine name for the gateway.
-// The name length is limited to 40 characters to ensure we don't hit the 63-character limit
+// The name length is limited to 20 characters to ensure we don't hit the 63-character limit
 // when generating the "machine public IP name".
-// At most 6 characters for the zone (which is usually very short),
-// at most 12 for the region and zone combined,
-// at most 32 for the infra id, region and zone combined
-// (the infra id is the longest significant piece of information here).
-// We add "-subgw-", 7 characters, for a total of 40 with the hyphen between region and zone.
-func MachineName(infraID, region, zone string) string {
-	if len(infraID)+len(region)+len(zone) > 32 {
-		// Limit the name length to 40 characters
-		if len(zone) > 6 {
-			zone = zone[0:6]
-		}
-
-		if len(region) > 12-len(zone) {
-			region = region[0 : 12-len(zone)]
-		}
-
-		if len(infraID) > 32-len(zone)-len(region) {
-			infraID = infraID[0 : 32-len(zone)-len(region)]
-		}
+// At most 7 characters for the region,
+// at most 13 for the region and a randomly generated UUID.
+// We add "subgw-", 6 characters, for a total of 20 with the hyphen between region and UUID.
+func MachineName(region string) string {
+	if len(region) > 7 {
+		region = region[0:7]
 	}
 
-	return infraID + submarinerGatewayGW + region + "-" + zone
+	return submarinerGatewayGW + region + "-" + string(uuid.NewUUID())[0:6]
 }
 
-func (d *ocpGatewayDeployer) getAvailabilityZones(gwNodes []v1.Node) (stringset.Interface, error) {
+func (d *ocpGatewayDeployer) getAvailabilityZones(gwNodes []unstructured.Unstructured) (stringset.Interface, error) {
 	zonesWithSubmarinerGW := stringset.New()
 
 	for i := range gwNodes {
-		zonesWithSubmarinerGW.Add(gwNodes[i].GetLabels()[topologyLabel])
+		zone, _, err := unstructured.NestedString(gwNodes[i].Object, "spec", "template", "spec", "providerSpec", "value", "zone")
+		if err != nil {
+			return nil, errors.Wrap(err, "error getting the zone from the existing node")
+		}
+
+		zonesWithSubmarinerGW.Add(zone)
 	}
 
 	ctx, cancel := context.WithTimeout(context.Background(), 300*time.Second)
@@ -360,9 +347,9 @@ func (d *ocpGatewayDeployer) Cleanup(status reporter.Interface) error {
 		return status.Error(err, "deleting gateway security group failed")
 	}
 
-	err = d.deleteGateway()
+	err = d.deleteGateway(status)
 	if err != nil {
-		return status.Error(err, "removing gateway node failed")
+		return err
 	}
 
 	status.Success("Removed gateway node")
@@ -370,13 +357,11 @@ func (d *ocpGatewayDeployer) Cleanup(status reporter.Interface) error {
 	return nil
 }
 
-func (d *ocpGatewayDeployer) deleteGateway() error {
-	gwNodes, err := d.azure.K8sClient.ListGatewayNodes()
+func (d *ocpGatewayDeployer) deleteGateway(status reporter.Interface) error {
+	machineSetList, err := d.msDeployer.List()
 	if err != nil {
-		return errors.Wrapf(err, "error getting the gw nodes")
+		return status.Error(err, "error listing the Submariner gateway nodes")
 	}
-
-	gwNodesList := gwNodes.Items
 
 	pubIPClient, err := d.getPublicIPClient()
 	if err != nil {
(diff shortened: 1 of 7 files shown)


pkg/azure/ocpgwdeployer.go
```diff
@@ -267,7 +267,7 @@ func (d *ocpGatewayDeployer) initMachineSet(name, zone string) (*unstructured.Un
 }
 
 func (d *ocpGatewayDeployer) deployGateway(zone string) error {
-	machineSet, err := d.initMachineSet(MachineName(d.azure.InfraID, d.azure.Region, zone), zone)
+	machineSet, err := d.initMachineSet(MachineName(d.azure.Region, zone), zone)
```
tpantelis: If we just need any valid, unique name for the machine set, why not just use a UUID or use the K8s generate name functionality?

aswinsuryan: the public IP will have this cluster name added to it. So the UUID should also be less than 20 characters[1].
Also when listing machineset wouldn't it be better to have a submariner identifier in it?

[1] https://github.com/openshift/machine-api-provider-azure/blob/main/pkg/cloud/azure/defaults.go#L86

pkg/azure/ocpgwdeployer.go
```diff
@@ -7,7 +7,7 @@ Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-http://www.apache.org/licenses/LICENSE-2.0
+    http://www.apache.org/licenses/LICENSE-2.0
```
skitt: Please don’t change this.

pkg/azure/ocpgwdeployer.go
```diff
@@ -276,15 +276,13 @@ func (d *ocpGatewayDeployer) deployGateway(zone string) error {
...
 // at most 12 for the region and zone combined,
-// at most 32 for the infra id, region and zone combined
-// (the infra id is the longest significant piece of information here).
-// We add "-subgw-", 7 characters, for a total of 40 with the hyphen between region and zone.
-func MachineName(infraID, region, zone string) string {
-	if len(infraID)+len(region)+len(zone) > 32 {
+// We add "-subgw-", 7 characters, for a total of 20 with the hyphen between region and zone.
+func MachineName(region, zone string) string {
+	if len(region)+len(zone) > 12 {
 		// Limit the name length to 40 characters
```
skitt: This comment needs to be updated too.

pkg/azure/ocpgwdeployer.go
```diff
@@ -318,7 +312,7 @@ func (d *ocpGatewayDeployer) getAvailabilityZones(gwNodes []v1.Node) (stringset.
 	}
 
 	pager := resourceSKUClient.NewListPager(&armcompute.ResourceSKUsClientListOptions{
-		Filter: pointer.String(d.azure.Region),
+		Filter: to.StringPtr(d.azure.Region),
```
skitt: go-autorest is deprecated and will no longer be supported soon, please don’t change this.

skitt (changes requested):

pkg/azure/ocpgwdeployer.go
aswinsuryan: I will revert,  I was just testing the PR. I think I did not raise the PR from the latest code.

pkg/azure/ocpgwdeployer.go
aswinsuryan: Done.

pkg/azure/ocpgwdeployer.go
aswinsuryan: Done. Sorry, it was not intentional.

pkg/azure/ocpgwdeployer.go
```diff
@@ -276,30 +276,24 @@ func (d *ocpGatewayDeployer) deployGateway(zone string) error {
...
 			region = region[0 : 12-len(zone)]
 		}
-
-		if len(infraID) > 32-len(zone)-len(region) {
-			infraID = infraID[0 : 32-len(zone)-len(region)]
-		}
 	}
 
-	return infraID + submarinerGatewayGW + region + "-" + zone
+	return submarinerGatewayGW + region + "-" + zone
```
sridhargaddam: I think this would create issues during upgrades. 

Say, we initially had an existing deployment where the Gatewaynode is created using a MachineSet. Following this PR if we change the MachineSet name, the reconciler code (if triggered from ACM or if user manually executes subctl cloud-prepare once again), will look for a MachineSet with a specific name and when it does not find it, will re-create it. Can you please explain how we handle this scenario?

pkg/azure/ocpgwdeployer.go
```diff
@@ -276,30 +276,24 @@ func (d *ocpGatewayDeployer) deployGateway(zone string) error {
...
-// (the infra id is the longest significant piece of information here).
-// We add "-subgw-", 7 characters, for a total of 40 with the hyphen between region and zone.
-func MachineName(infraID, region, zone string) string {
-	if len(infraID)+len(region)+len(zone) > 32 {
-		// Limit the name length to 40 characters
-		if len(zone) > 6 {
-			zone = zone[0:6]
+// We add "subgw-", 6 characters, for a total of 20 with the hyphen between region and zone.
+func MachineName(region, zone string) string {
+	if len(region)+len(zone) > 12 {
```
sridhargaddam: If we are using the region/zone + some submariner identifier in the names, it means we can only create a limited number of Gateway nodes and cannot scale them beyond the number of zones. I think this is not specific to this PR/Azure but is applicable to all the other cloud-prepare methods.

Ideally, I feel we should use some UUID for the names (with a suitable length) and label (or tag) the MachineSetObjects to identify the Submariner resources rather than creating names with some pattern which is error prone.

sridhargaddam (changes requested):

pkg/azure/ocpgwdeployer.go
aswinsuryan: We list the g/w nodes [1] using the submariner g/w label and if we have a sufficient number of nodes we skip the creation. For deletion we use the same API and check for the substring subgw- (earlier it was -subgw-) and it should work as well, 

[1]https://github.com/submariner-io/cloud-prepare/blob/devel/pkg/azure/ocpgwdeployer.go#L99

pkg/azure/ocpgwdeployer.go
aswinsuryan: we decided to use Zone assuming we will have one g/w per availability zone. We follow this for all cloud expect Openstack. Can we raise a separate issue and track it over there?

sridhargaddam: Yes, please report an issue for this. Thanks.

pkg/azure/ocpgwdeployer.go
sridhargaddam: I see that we are not using this API or the submarinerGatewayGW as part of delete*, but identify the machineSet/prefix/name using the "-" char. IMHO this is fragile. Can you please fix it or add a comment in the cleanup path about this dependency.

* https://github.com/submariner-io/cloud-prepare/blob/4b2f851014e5b98b47cd7c69e36dcc4e62b245f8/pkg/azure/ocpgwdeployer.go#L403

sridhargaddam (changes requested):

maayanf24:
@aswinsuryan - please add release notes

pkg/azure/ocpgwdeployer.go
aswinsuryan: https://github.com/submariner-io/cloud-prepare/issues/532

pkg/azure/ocpgwdeployer.go
aswinsuryan: I will modify the delete code and fill include the fix for multiple g/w nodes getting spawned in Azure cluster.

aswinsuryan: Updated the delete logic.

pkg/azure/ocpgwdeployer.go
```diff
@@ -360,7 +349,7 @@ func (d *ocpGatewayDeployer) Cleanup(status reporter.Interface) error {
 		return status.Error(err, "deleting gateway security group failed")
 	}
 
-	err = d.deleteGateway()
+	err = d.deleteGateway(status)
 	if err != nil {
 		return status.Error(err, "removing gateway node failed")
```
tpantelis: `deleteGateway` now logs errors to the `status` so we should just return the error here.

pkg/azure/ocpgwdeployer.go
```diff
@@ -386,32 +373,81 @@ func (d *ocpGatewayDeployer) deleteGateway() error {
...
 	defer cancel()
 
-	for i := 0; i < len(gwNodesList); i++ {
-		if !strings.Contains(gwNodesList[i].Name, submarinerGatewayGW) {
-			err = d.K8sClient.RemoveGWLabelFromWorkerNode(&gwNodesList[i])
-			if err != nil {
-				return errors.Wrapf(err, "failed to remove labels from worker node")
-			}
+	for i := range machineSetList {
+		status.Start(fmt.Sprintf("Deleting the gateway instance %q", machineSetList[i].GetName()))
```
tpantelis:
```suggestion
		status.Start("Deleting the gateway instance %q", machineSetList[i].GetName())
```

pkg/azure/ocpgwdeployer.go
```diff
@@ -386,32 +373,81 @@ func (d *ocpGatewayDeployer) deleteGateway() error {
...
 
-			err = d.deletePublicIP(ctx, pubIPClient, publicIPName)
-			if err != nil {
-				return errors.Wrapf(err, "failed to delete public-ip")
-			}
+		publicIPName := machineSetList[i].GetName() + "-pub"
+
+		err = d.deletePublicIP(ctx, pubIPClient, publicIPName)
+		if err != nil {
+			return errors.Wrapf(err, "failed to delete public-ip")
```
tpantelis:
```suggestion
			return status.Error(err, "failed to delete public-ip %q", publicIPName)
```

pkg/azure/ocpgwdeployer.go
```diff
@@ -386,32 +373,81 @@ func (d *ocpGatewayDeployer) deleteGateway() error {
...
-			zone := machineSetName[strings.LastIndex(gwNodesList[i].Name, "-")-1:]
+			err = d.K8sClient.RemoveGWLabelFromWorkerNode(&gwNodes[i])
+		}
 
-			machineSet, err := d.initMachineSet(prefix, zone)
-			if err != nil {
-				return err
-			}
+		if err != nil {
+			return errors.Wrapf(err, "failed to cleanup node %q", gwNodes[i].Name)
```
tpantelis: We should be consistent returning `status.Error` vs `errors.Wrapf.`

```suggestion
			return status.Error(err, "failed to cleanup node %q", gwNodes[i].Name)
```

pkg/azure/ocpgwdeployer.go
```diff
@@ -386,32 +373,81 @@ func (d *ocpGatewayDeployer) deleteGateway() error {
...
+		}
+
+		publicIPName := gwNodes[i].Name + "-pub"
 
-			return errors.Wrapf(d.msDeployer.Delete(machineSet), "error deleting machine set %q", machineSet.GetName())
+		err = d.deletePublicIP(ctx, pubIPClient, publicIPName)
+		if err != nil {
+			return errors.Wrapf(err, "failed to delete public-ip")
```
tpantelis:
```suggestion
			return status.Error(err, "failed to delete public-ip")
```

pkg/azure/ocpgwdeployer.go
```diff
@@ -79,40 +79,36 @@ func (d *ocpGatewayDeployer) Deploy(input api.GatewayDeployInput, status reporte
 
 	status.Start("Deploying gateway node")
 
-	nsgClient, err := d.getNsgClient()
+	nsgClient, nwClient, pubIPClient, err := d.getClients(status)
 	if err != nil {
-		return status.Error(err, "Failed to get network security groups client")
+		return status.Error(err, "Failed to get client")
```
tpantelis: `getClients` logs errors.

```suggestion
		return err
```

pkg/azure/ocpgwdeployer.go
```diff
@@ -79,40 +79,36 @@ func (d *ocpGatewayDeployer) Deploy(input api.GatewayDeployInput, status reporte
...
-	if err != nil {
-		return status.Error(err, "Failed to get network interfaces client")
-	}
+	groupName := d.InfraID + externalSecurityGroupSuffix
 
-	pubIPClient, err := d.getPublicIPClient()
+	machineSets, err := d.msDeployer.List()
 	if err != nil {
-		return status.Error(err, "Failed to get network public IP addresses client")
+		return errors.Wrap(err, "error getting the gateway machinesets")
```
tpantelis: For consistency...

```suggestion
		return status.Error(err, "error getting the gateway machinesets")
```

pkg/k8s/k8siface.go
```diff
@@ -126,3 +128,15 @@ func (k *k8sIface) RemoveGWLabelFromWorkerNode(node *v1.Node) error {
 		delete(existing.Labels, SubmarinerGatewayLabel)
 	})
 }
+
+func (k *k8sIface) FilterDedicatedNode(gwNodes []v1.Node) []v1.Node {
+	var resultNode []v1.Node
+
+	for i := 0; i < len(gwNodes); i++ {
+		if !(gwNodes[i].GetLabels()[ocp.MachineSetOwnerLabel] == "submariner") {
```
tpantelis: Easier to read...

```suggestion
		if gwNodes[i].GetLabels()[ocp.MachineSetOwnerLabel] != "submariner" {
```

pkg/k8s/k8siface.go
```diff
@@ -126,3 +128,15 @@ func (k *k8sIface) RemoveGWLabelFromWorkerNode(node *v1.Node) error {
 		delete(existing.Labels, SubmarinerGatewayLabel)
 	})
 }
+
+func (k *k8sIface) FilterDedicatedNode(gwNodes []v1.Node) []v1.Node {
```
tpantelis: This API is a wrapper for K8s so this function doesn't seem to belong here. It is tied to machine sets  and references the `ocp` package and so it seems better suited there, ie `ocp.FilterDedicatedNodes`.

pkg/ocp/machinesets_test.go
```diff
@@ -193,22 +193,34 @@ var _ = Describe("K8s MachineSetDeployer", func() {
...
+		"kind":       "MachineSet",
+		"metadata": map[string]interface{}{
+			"namespace": "test-ns",
+		},
+		"spec": map[string]interface{}{
+			"template": map[string]interface{}{
+				"spec": map[string]interface{}{
+					"metadata": map[string]interface{}{
+						"labels": map[string]interface{}{
+							"submariner.io/machine-set-owner": owner,
```
tpantelis: Use constant here.

pkg/rhos/ocpgwdeployer.go
```diff
@@ -156,11 +155,18 @@ func (d *ocpGatewayDeployer) Deploy(input api.GatewayDeployInput, status reporte
 		return status.Error(err, "creating gateway security group failed")
 	}
 
+	machineSets, err := d.msDeployer.List()
+	if err != nil {
+		return errors.Wrap(err, "error getting the gateway machinesets")
```
tpantelis:
```suggestion
		return status.Error(err, "error getting the gateway machinesets")
```

pkg/rhos/ocpgwdeployer.go
```diff
@@ -156,11 +155,18 @@ func (d *ocpGatewayDeployer) Deploy(input api.GatewayDeployInput, status reporte
...
+	}
+
 	gwNodes, err := d.K8sClient.ListGatewayNodes()
 	if err != nil {
-		return status.Error(err, "listing the existing gateway nodes failed")
+		return errors.Wrap(err, "listing the existing gateway nodes failed")
```
tpantelis:
```suggestion
		return status.Error(err, "listing the existing gateway nodes failed")
```

pkg/rhos/ocpgwdeployer.go
```diff
@@ -316,11 +316,7 @@ func (d *ocpGatewayDeployer) Cleanup(status reporter.Interface) error {
 	gwNodes := gwNodesList.Items
 
 	for i := range gwNodes {
-		// Check if the instance belongs to the cluster (identified via infraID) we are operating on.
-		// If the instance name matches with d.InfraID + "-submariner-gw-", it implies that
-		// the gateway node was deployed using the OCPMachineSet which is already handled above and
-		// the node should be already be in deleting state.
-		if !strings.HasPrefix(gwNodes[i].Name, d.InfraID) || strings.HasPrefix(gwNodes[i].Name, prefix) {
+		if gwNodes[i].GetLabels()[ocp.MachineSetOwnerLabel] == "submariner" {
```
tpantelis: Should make "submariner" a constant in `ocp` as it's referenced in several places, eg `LabelValueOwnedBy`.

tpantelis (changes requested):

aswinsuryan:
@sridhargaddam @tpantelis  Each vm is created(will have same name as machine set) in its own resource group. So having a same name vm in different clusters should not be an issue. Should we still change vm name to just uuid?

tpantelis (approved):

pkg/azure/gw-machineset.go
```diff
@@ -44,6 +44,7 @@ spec:
       metadata:
         labels:
           submariner.io/gateway: "true"
+          submariner.io/machine-set-owner: "submariner"
```
sridhargaddam: As discussed on the call, lets get rid of this new label as its introducing lot of corner cases.

sridhargaddam (changes requested):
Thanks for the PR @aswinsuryan

pkg/azure/gw-machineset.go
aswinsuryan: Done.

pkg/ocp/machinesets.go
```diff
@@ -207,19 +206,28 @@ func (msd *k8sMachineSetDeployer) List() ([]unstructured.Unstructured, error) {
...
-func FilterDedicatedNode(gwNodes []v1.Node) []v1.Node {
+func RemoveDuplicates(machineSets []unstructured.Unstructured, gwNodes []v1.Node) []v1.Node {
 	var resultNode []v1.Node
 
 	for i := 0; i < len(gwNodes); i++ {
-		if gwNodes[i].GetLabels()[MachineSetOwnerLabel] != MachineSetOwnerSubmariner {
+		addToResult := true
+
+		for i := 0; i < len(machineSets); i++ {
+			if strings.Contains(gwNodes[i].GetName(), machineSets[i].GetName()) {
```
sridhargaddam, line 224: Can we safely assume that machineSet.name will be same as gwNode.Name?

Following is the output from one of the deployments.

```
[sgaddam@localhost submariner]$ oc get nodes
NAME         STATUS   ROLES                  AGE   VERSION
master-0-0   Ready    control-plane,master   19d   v1.25.4+77bec7a
master-0-1   Ready    control-plane,master   19d   v1.25.4+77bec7a
master-0-2   Ready    control-plane,master   19d   v1.25.4+77bec7a
worker-0-0   Ready    worker                 19d   v1.25.4+77bec7a
worker-0-1   Ready    worker                 19d   v1.25.4+77bec7a

```

```
[sgaddam@localhost submariner]$ oc get machinesets -A
NAMESPACE               NAME                     DESIRED   CURRENT   READY   AVAILABLE   AGE
openshift-machine-api   ice01-0-qjgw7-worker-0   2         2         2       2           19d

```

aswinsuryan: I think it should depend on the machineset provider, for Azure and RHOS that is the pattern I observed. We may need an extra label, if we are planning not using the name.

sridhargaddam: Even if its a pattern today, things might change anytime. So, IMHO we should identify alternatives to this.

Please check if there is a way to know the node/machine-name from the machineset - I think there should be one (the machineset controller should be using it to identify the associated machines).

aswinsuryan: Just checked, 

the below one to get info about node from machineset return emty

`% kubectl get machinesets -n openshift-machine-api subgw-eastus-05ea1c -o yaml | grep subgw-eastus-05ea1c-6475l`

and when tried the otherway , it does not have any reference to the machine name.

```
% kubectl get nodes subgw-eastus-05ea1c-6475l -o yaml | grep  subgw-eastus-05ea1c                                         
    cloud.network.openshift.io/egress-ipconfig: '[{"interface":"subgw-eastus-05ea1c-6475l-nic","ifaddr":{"ipv4":"10.0.128.0/17"},"capacity":{"ip":255}}]'
    csi.volume.kubernetes.io/nodeid: '{"disk.csi.azure.com":"subgw-eastus-05ea1c-6475l","file.csi.azure.com":"subgw-eastus-05ea1c-6475l"}'
    machine.openshift.io/machine: openshift-machine-api/subgw-eastus-05ea1c-6475l
    kubernetes.io/hostname: subgw-eastus-05ea1c-6475l
  name: subgw-eastus-05ea1c-6475l
  providerID: azure:///subscriptions/3c7a4fa5-8496-4cdc-8a78-e344e2f896d7/resourceGroups/asuryana1-tbrrj-rg/providers/Microsoft.Compute/virtualMachines/subgw-eastus-05ea1c-6475l
  - address: subgw-eastus-05ea1c-6475l
```

sridhargaddam: Okay, I see that we have an annotation on the nodes as shown below. But this may not be any different to the assumption we are making about the names. 

```
machine.openshift.io/machine: openshift-machine-api/ice01-1-pgnls-master-0
```

pkg/azure/ocpgwdeployer.go
```diff
@@ -96,7 +95,7 @@ func (d *ocpGatewayDeployer) Deploy(input api.GatewayDeployInput, status reporte
 	}
 
 	gwNodeItems := gwNodes.Items
-	taggedExistingNodes := ocp.FilterDedicatedNode(gwNodeItems)
+	taggedExistingNodes := ocp.RemoveDuplicates(nil, gwNodeItems)
```
sridhargaddam: why nil?

aswinsuryan: It was an error while refactoring. Fixed it.

pkg/ocp/machinesets_test.go
```diff
@@ -176,7 +176,7 @@ var _ = Describe("K8s MachineSetDeployer", func() {
 				machineSet.SetName(machineSetName)
 				_, err := msClient.Create(context.TODO(), machineSet, metav1.CreateOptions{})
 				Expect(err).To(Succeed())
-				machineSet = newMachineSet("other")
+				machineSet = newMachineSet("fasle")
```
sridhargaddam: Is this typo or intentional?

sridhargaddam (changes requested):

pkg/ocp/machinesets_test.go
aswinsuryan: no .. :D

aswinsuryan: Done

sridhargaddam (approved):

sridhargaddam:
@aswinsuryan please update the commit message in the PR.

aswinsuryan:
> @aswinsuryan please update the commit message in the PR.

@sridhargaddam done.

skitt (approved):

PR #571, opened 2023-01-26
Simplify API by exposing `OpenPorts`

mkolesnik:
Since our "cloud preparation" is only opening ports (and not expected to change), simplify the API to make it clear that's what it's doing.

The `PrepareForSubmariner` function was initially intended to be used to do all internal & external preparations, but as gateway deployment has been split out (quite a long time ago), it only opens internal ports now.

The `PrepareForSubmariner` function is marked as deprecated and will be removed on the 0.16 release.


pkg/api/api.go
```diff
@@ -33,9 +33,12 @@ type PrepareForSubmarinerInput struct {
...
 type Cloud interface {
-	// PrepareForSubmariner will prepare the cloud for Submariner to operate on.
+	// Deprecated: Use OpenPorts instead, this will be removed on the 0.16 release.
 	PrepareForSubmariner(input PrepareForSubmarinerInput, status reporter.Interface) error
 
+	// OpenPorts inside the cloud for submariner to communicate through.
+	OpenPorts(ports []PortSpec, status reporter.Interface) error
+
 	// CleanupAfterSubmariner will clean up the cloud after Submariner is removed.
 	CleanupAfterSubmariner(status reporter.Interface) error
```
skitt: Should this be replaced by a `ClosePorts` while we’re at it?

skitt (changes requested):
Please update `README.md` too.

pkg/api/api.go
mkolesnik: I was thinking of it, could do if the direction seems good

pkg/azure/azure.go
```diff
@@ -38,19 +38,22 @@ func NewCloud(info *CloudInfo) api.Cloud {
 }
 
 func (az *azureCloud) PrepareForSubmariner(input api.PrepareForSubmarinerInput, reporter reporterInterface.Interface) error {
+	return az.OpenPorts(input.InternalPorts, reporter)
+}
+
+func (az *azureCloud) OpenPorts(ports []api.PortSpec, reporter reporterInterface.Interface) error {
 	reporter.Start("Opening internal ports for intra-cluster communications on Azure")
```
skitt, line 41: Even with the API simplification, this should still avoid causing errors when called with no ports:

```suggestion
	if (len(ports) == 0) {
		return nil
	}
	
	reporter.Start("Opening internal ports for intra-cluster communications on Azure")
```

The same applies to the other implementations.

pkg/api/api.go
```diff
@@ -33,9 +33,12 @@ type PrepareForSubmarinerInput struct {
 
 // Cloud is a potential cloud for installing Submariner on.
 type Cloud interface {
-	// PrepareForSubmariner will prepare the cloud for Submariner to operate on.
+	// Deprecated: Use OpenPorts instead, this will be removed on the 0.16 release.
```
tpantelis: I think we can just remove this function now rather than adding tech debt. Users of the API can just adjust when they bump the go mod version (they'll have to adjust at some point anyway). There's really only 2 users of this API and we control both anyway.

pkg/azure/azure.go
mkolesnik: IDK if I support this position, would you call a command that does something with not enough arguments and expect it to silently succeed? Not usually, unless you use some`-f` flag.

If anything, this should return an error.

pkg/api/api.go
mkolesnik: I'm up for it if there's consensus to do so, would certainly be cleaner.

pkg/azure/azure.go
tpantelis: I agree with @skitt. It's better to handle the trivial case here rather than forcing all users to do it or risk a latent bug.

mkolesnik: Why would it be acceptable in this case to send an empty list? It should be an error

tpantelis: Somebody has to verify/handle an empty list - it might as well be at the source rather than expecting all callers to check for it - makes it a simpler API to use. I don't consider it an error - the API is to open the given ports - the trivial case is that the list happens to be empty in which case it's a noop. There's many similar trivial cases in other APIs, eg a sort function where it simply returns if the input list is empty instead of crashing or throwing an error.

mkolesnik: Regardless of the consensus here, its not part of this change, I'll send a separate PR with a proposal for this specifically.

mkolesnik: Please see #580

skitt (approved):

tpantelis (approved):

PR #580, opened 2023-02-01
Warn and return when no ports given to `OpenPorts`

mkolesnik:
As the operation for opening ports involves multiple communications with the cloud, and various necessary permissions, make sure that when no ports are given we don't try to do it but instead warn the user about it (as generally, callers should be wary of calling the API with no ports).

Depends on #571 

### DIFF
--- a/pkg/aws/aws.go
+++ b/pkg/aws/aws.go
@@ -88,7 +88,7 @@ func DefaultProfile() string {
 	return "default"
 }
 
-func (ac *awsCloud) PrepareForSubmariner(input api.PrepareForSubmarinerInput, status reporter.Interface) error {
+func (ac *awsCloud) OpenPorts(ports []api.PortSpec, status reporter.Interface) error {
 	status.Start(messageRetrieveVPCID)
 	defer status.End()
 
@@ -108,7 +108,7 @@ func (ac *awsCloud) PrepareForSubmariner(input api.PrepareForSubmarinerInput, st
 
 	status.Success(messageValidatedPrerequisites)
 
-	for _, port := range input.InternalPorts {
+	for _, port := range ports {
 		status.Start("Opening port %v protocol %s for intra-cluster communications", port.Port, port.Protocol)
 
 		err = ac.allowPortInCluster(vpcID, port.Port, port.Protocol)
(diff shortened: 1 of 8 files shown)


pkg/aws/aws.go
```diff
@@ -89,9 +89,15 @@ func DefaultProfile() string {
 }
 
 func (ac *awsCloud) OpenPorts(ports []api.PortSpec, status reporter.Interface) error {
-	status.Start(messageRetrieveVPCID)
 	defer status.End()
 
+	if len(ports) == 0 {
+		status.Warning("Ignoring attempt to open ports with no ports given")
```
skitt, line 95: Warnings shown to the user should be actionable; this isn’t something the user can fix.

skitt (changes requested):

pkg/aws/aws.go
mkolesnik: Hence I wanted it to return an error, but there was pushback against it even thought it is an unintended behavior.

This was a compromise I was trying to suggest, what do you suggest?

skitt: I suggest returning directly:

```suggestion
```

mkolesnik: Calling "open ports" with no ports is probably some bug on the client, hence I think it would be better to either warn or even error out in such a case.

I prefer not hiding possible bugs by silently succeeding.

tpantelis: > I prefer not hiding possible bugs by silently succeeding.

I understand your point but it's not necessarily a bug on the client, in fact I'd say it probably isn't in this case. As I've mentioned before, there's many APIs out there that silently handle the trivial case of no input/nothing to do, in fact it's usually a convenience for callers.  That said this isn't a widely-used API so I don't think it really matters either way so I'm not going to engage in a battle over this 😄

skitt: My main point is that this isn’t great UX: we foist the warning on the user, even though they can’t do anything about it. What do users do when they get a warning they don’t understand? We can hope they’ll ignore it if everything works ultimately, but if that’s what we’re hoping for, we might as well not show the warning.

As far as this being a bug, my preference is to handle the empty slice in one place (here), and get rid of all the `if`s elsewhere. If we do that, then this isn’t a bug.

PR #594, opened 2023-02-08
For Azure cloud prepare get the image from existing Machineset

aswinsuryan:
For Azure cloud prepare it uses a hard-coded path and infra id for the machine set. Now it is changed the image. from one of the existing worker nodes.

Signed-off-by: Aswin Suryanarayanan <aswinsuryan@gmail.com>

### DIFF
--- a/pkg/azure/ocpgwdeployer.go
+++ b/pkg/azure/ocpgwdeployer.go
@@ -127,7 +127,7 @@ func (d *ocpGatewayDeployer) Deploy(input api.GatewayDeployInput, status reporte
 	if d.dedicatedGWNode {
 		image, imageErr := d.msDeployer.GetWorkerNodeImage(nil, nil, d.InfraID)
 		if imageErr != nil {
-			return errors.Wrap(err, "error retrieving worker node image")
+			return errors.Wrap(imageErr, "error retrieving worker node image")
 		}
 
 		err = d.deployDedicatedGWNode(machineSets, gatewayNodesToDeploy, image, status)
--- a/pkg/azure/ocpgwdeployer.go
+++ b/pkg/azure/ocpgwdeployer.go
@@ -125,7 +125,12 @@ func (d *ocpGatewayDeployer) Deploy(input api.GatewayDeployInput, status reporte
 	}
 
 	if d.dedicatedGWNode {
-		err = d.deployDedicatedGWNode(machineSets, gatewayNodesToDeploy, status)
+		image, imageErr := d.msDeployer.GetWorkerNodeImage(nil, nil, d.InfraID)
+		if imageErr != nil {
+			return errors.Wrap(err, "error retrieving worker node image")
+		}
+
+		err = d.deployDedicatedGWNode(machineSets, gatewayNodesToDeploy, image, status)
 	} else {
 		err = d.tagExistingNode(nsgClient, nwClient, pubIPClient, gatewayNodesToDeploy, status)
 	}
@@ -138,7 +143,7 @@ func (d *ocpGatewayDeployer) Deploy(input api.GatewayDeployInput, status reporte
 }
 
 func (d *ocpGatewayDeployer) deployDedicatedGWNode(gwNodes []unstructured.Unstructured, gatewayNodesToDeploy int,
-	status reporter.Interface,
+	image string, status reporter.Interface,
 ) error {
 	az, err := d.getAvailabilityZones(gwNodes)
 	if err != nil || az.Size() == 0 {
@@ -148,7 +153,7 @@ func (d *ocpGatewayDeployer) deployDedicatedGWNode(gwNodes []unstructured.Unstru
 	for _, zone := range az.Elements() {
 		status.Start("Deploying dedicated gateway node")
 
-		err := d.deployGateway(zone)
+		err := d.deployGateway(zone, image)
 		if err != nil {
 			return status.Error(err, "error deploying gateway for zone %q", zone)
 		}
@@ -217,9 +222,10 @@ type machineSetConfig struct {
 	InfraID      string
 	InstanceType string
 	Region       string
+	Image        string
 }
 
-func (d *ocpGatewayDeployer) loadGatewayYAML(name, zone string) ([]byte, error) {
+func (d *ocpGatewayDeployer) loadGatewayYAML(name, zone, image string) ([]byte, error) {
 	var buf bytes.Buffer
 
 	tpl, err := template.New("").Parse(machineSetYAML)
@@ -233,6 +239,7 @@ func (d *ocpGatewayDeployer) loadGatewayYAML(name, zone string) ([]byte, error) 
 		InstanceType: d.instanceType,
 		Region:       d.azure.Region,
 		AZ:           zone,
+		Image:        image,
 	}
 
 	err = tpl.Execute(&buf, tplVars)
@@ -243,8 +250,8 @@ func (d *ocpGatewayDeployer) loadGatewayYAML(name, zone string) ([]byte, error) 
 	return buf.Bytes(), nil
 }
 
-func (d *ocpGatewayDeployer) initMachineSet(name, zone string) (*unstructured.Unstructured, error) {
-	gatewayYAML, err := d.loadGatewayYAML(name, zone)
+func (d *ocpGatewayDeployer) initMachineSet(name, zone, image string) (*unstructured.Unstructured, error) {
+	gatewayYAML, err := d.loadGatewayYAML(name, zone, image)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


nyechiel:
Setting priority as high and marking for backport, as currently all OpenShift 4.12 Azure based deployments are blocked without this fix.

pkg/azure/ocpgwdeployer.go
```diff
@@ -436,3 +446,30 @@ func (d *ocpGatewayDeployer) getClients(status reporter.Interface) (
 
 	return nsgClient, nwClient, pubIPClient, nil
 }
+
+func (d *ocpGatewayDeployer) getWorkerNodeImage() (string, error) {
```
skitt: This could be defined in `ocp/machinesets.go`, which would avoid exposing `ClientForMsd()`.

skitt (dismissed):

pkg/azure/ocpgwdeployer.go
aswinsuryan: Done.

pkg/ocp/machinesets.go
```diff
@@ -132,25 +139,41 @@ func (msd *k8sMachineSetDeployer) GetWorkerNodeImage(workerNodeList []string, ma
 			}
 		}
 
-		disks, _, _ := unstructured.NestedSlice(existing.Object, "spec", "template", "spec", "providerSpec", "value", "disks")
-		if len(disks) == 0 {
-			image, _, _ := unstructured.NestedString(existing.Object, "spec", "template", "spec", "providerSpec", "value", "image")
+		image := getImageFromMachineSet(existing)
+		if image != "" {
```
mkolesnik:
```suggestion
		if image := getImageFromMachineSet(existing); image != "" {
```

pkg/ocp/machinesets.go
```diff
@@ -132,25 +139,41 @@ func (msd *k8sMachineSetDeployer) GetWorkerNodeImage(workerNodeList []string, ma
...
+			return image, nil
+		}
+	}
+
+	return "", fmt.Errorf("could not retrieve the image of one of the worker nodes from the infra %q", infraID)
+}
+
+func getImageFromMachineSet(existing *unstructured.Unstructured) string {
+	disks, _, _ := unstructured.NestedSlice(existing.Object, "spec", "template", "spec", "providerSpec", "value", "disks")
+	if len(disks) == 0 {
```
mkolesnik:
```suggestion
```

The whole if/else structure can just be dropped and make the function simpler.
If you found disks and an image there, use it, otherwise keep searching. 

The more correct way would've been to check the `spec.template.spec.providerSpec.kind` field and only search for the expected image field based off that, but I think the current code, without the `if len(disks) == 0 {`, is good enough.

pkg/ocp/machinesets.go
```diff
@@ -98,9 +98,16 @@ func (msd *k8sMachineSetDeployer) clientForMsd(nameSpace string) dynamic.Resourc
...
 	infraID string,
 ) (string, error) {
-	machineSetClient, err := msd.clientFor(machineSet)
-	if err != nil {
-		return "", err
+	var machineSetClient dynamic.ResourceInterface
+
+	if machineSet == nil {
+		machineSetClient = msd.clientForMsd("openshift-machine-api")
+	} else {
```
mkolesnik:
```suggestion
	machineSetClient := msd.clientForMsd("openshift-machine-api")

	if machineSet != nil {
```

pkg/ocp/machinesets.go
```diff
@@ -98,9 +98,16 @@ func (msd *k8sMachineSetDeployer) clientForMsd(nameSpace string) dynamic.Resourc
...
+		var err error
+		machineSetClient, err = msd.clientFor(machineSet)
```
mkolesnik:
```suggestion
		machineSetClient, err := msd.clientFor(machineSet)
```
?

pkg/ocp/machinesets.go
skitt: If you like that style it can be applied to the previous two `if`s as well ;-).

pkg/ocp/machinesets.go
aswinsuryan: Done.

pkg/ocp/machinesets.go
aswinsuryan: Done.

pkg/ocp/machinesets.go
aswinsuryan: This was leading to an error "Unused variable", since it was already defined above after refactoring.

pkg/ocp/machinesets.go
aswinsuryan: Done, wherever possible

pkg/azure/ocpgwdeployer.go
```diff
@@ -125,7 +125,12 @@ func (d *ocpGatewayDeployer) Deploy(input api.GatewayDeployInput, status reporte
 	}
 
 	if d.dedicatedGWNode {
-		err = d.deployDedicatedGWNode(machineSets, gatewayNodesToDeploy, status)
+		image, imageErr := d.msDeployer.GetWorkerNodeImage(nil, nil, d.InfraID)
+		if imageErr != nil {
```
skitt, line 129:
```suggestion
		image, err := d.msDeployer.GetWorkerNodeImage(nil, nil, d.InfraID)
		if err != nil {
```

skitt (approved):
Minor idiomatic change, but not worth respinning. I have another concern regarding GetWorkerNodeImage but it’s not specific to this PR, I’ll submit another PR for it.

pkg/azure/ocpgwdeployer.go
aswinsuryan: used a different name since line  135  uses the err variable(declared outside this block) and its status is checked outside this block at line 138.

mkolesnik: Why does it matter though?

You return at next line, so it will never reach any code outside the blog.

Not that it matters that much, but you can use `err` here which is more straightforward.

tpantelis: yeah I agree with `err` here - it's local and doesn't escape the scope of this block. Also it needs to match L130 below which it doesn't.

pkg/ocp/machinesets.go
```diff
@@ -124,33 +130,47 @@ func (msd *k8sMachineSetDeployer) GetWorkerNodeImage(workerNodeList []string, ma
...
+			return image, nil
+		}
+	}
+
+	return "", fmt.Errorf("could not retrieve the image of one of the worker nodes from the infra %q", infraID)
+}
+
+func getImageFromMachineSet(existing *unstructured.Unstructured) string {
+	disks, _, _ := unstructured.NestedSlice(existing.Object, "spec", "template", "spec", "providerSpec", "value", "disks")
+	if len(disks) != 0 {
```
mkolesnik:
```suggestion
```

This can be dropped, golang will skip iterating over nil/empty slice

mkolesnik (approved):
Generally looks good, one suggestion if youre going to respin the PR, otherwise don't bother

pkg/azure/ocpgwdeployer.go
```diff
@@ -125,7 +125,12 @@ func (d *ocpGatewayDeployer) Deploy(input api.GatewayDeployInput, status reporte
...
+			return errors.Wrap(err, "error retrieving worker node image")
```
skitt:
```suggestion
		image, err := d.msDeployer.GetWorkerNodeImage(nil, nil, d.InfraID)
		if err != nil {
			return errors.Wrap(err, "error retrieving worker node image")
```

skitt (changes requested):
This hasn’t been resolved (and I’d missed the mismatch in line 130, which means it needs to be resolved).

skitt (dismissed):

pkg/azure/ocpgwdeployer.go
aswinsuryan: This lead to linting error 

pkg/azure/ocpgwdeployer.go:133:3: ineffectual assignment to err (ineffassign)
		err = d.deployDedicatedGWNode(machineSets, gatewayNodesToDeploy, image, status)

Hence reverting, Will fix the issue in  line 132.

pkg/azure/ocpgwdeployer.go
```diff
@@ -125,7 +125,12 @@ func (d *ocpGatewayDeployer) Deploy(input api.GatewayDeployInput, status reporte
 	}
 
 	if d.dedicatedGWNode {
-		err = d.deployDedicatedGWNode(machineSets, gatewayNodesToDeploy, status)
+		image, imageErr := d.msDeployer.GetWorkerNodeImage(nil, nil, d.InfraID)
+		if imageErr != nil {
+			return errors.Wrap(imageErr, "error retrieving worker node image")
+		}
```
skitt, line 131:
```suggestion
		var image string
		
		image, err = d.msDeployer.GetWorkerNodeImage(nil, nil, d.InfraID)
		if err != nil {
			return errors.Wrap(err, "error retrieving worker node image")
		}
```

skitt (approved):

skitt:
@aswinsuryan can you take care of the 0.13 and 0.12 backports?

aswinsuryan:
> @aswinsuryan can you take care of the 0.13 and 0.12 backports?

It requires two other back ports to resolve conflicts, 
Below is the  first one 
https://github.com/submariner-io/cloud-prepare/pull/597

It is not required in 0.12 as Azure support is added in 0.13

https://submariner.io/community/releases/#v0130

PR #595, opened 2023-02-10
Avoid double MachineSet retrieval

skitt:
GetWorkerNodeImage, if given an empty list of worker nodes, retrieves them all to extract their names, then retrieves them all *again* to process them. Avoid this by listing the nodes once, and processing that list.

### DIFF
--- a/pkg/ocp/machinesets.go
+++ b/pkg/ocp/machinesets.go
@@ -95,7 +95,7 @@ func (msd *k8sMachineSetDeployer) clientForMsd(nameSpace string) dynamic.Resourc
 	return msd.dynamicClient.Resource(machinesetGVR).Namespace(nameSpace)
 }
 
-func (msd *k8sMachineSetDeployer) GetWorkerNodeImage(workerNodeList []string, machineSet *unstructured.Unstructured,
+func (msd *k8sMachineSetDeployer) GetWorkerNodeImage(workerNodeNames []string, machineSet *unstructured.Unstructured,
 	infraID string,
 ) (string, error) {
 	machineSetClient := msd.clientForMsd("openshift-machine-api")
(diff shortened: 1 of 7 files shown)


pkg/ocp/machinesets.go
```diff
@@ -95,46 +95,51 @@ func (msd *k8sMachineSetDeployer) clientForMsd(nameSpace string) dynamic.Resourc
...
+	} else {
+		workerNodes = make([]unstructured.Unstructured, len(workerNodeNames))
 
-	for _, nodeName := range workerNodeList {
-		existing, err := machineSetClient.Get(context.TODO(), nodeName, metav1.GetOptions{})
-		if apierrors.IsNotFound(err) {
-			continue
-		}
+		for _, workerNodeName := range workerNodeNames {
+			workerNode, err := machineSetClient.Get(context.TODO(), workerNodeName, metav1.GetOptions{})
```
skitt: This could be improved if field selectors supports “in” searches, but as far as I’m aware they only support exact matches.

dfarrell07 (approved):

tpantelis (approved):

pkg/ocp/machinesets.go
```diff
@@ -95,7 +95,7 @@ func (msd *k8sMachineSetDeployer) clientForMsd(nameSpace string) dynamic.Resourc
 	return msd.dynamicClient.Resource(machinesetGVR).Namespace(nameSpace)
 }
 
-func (msd *k8sMachineSetDeployer) GetWorkerNodeImage(workerNodeList []string, machineSet *unstructured.Unstructured,
+func (msd *k8sMachineSetDeployer) GetWorkerNodeImage(workerNodeNames []string, machineSet *unstructured.Unstructured,
```
mkolesnik:
```suggestion
func (msd *k8sMachineSetDeployer) GetWorkerNodeImage(machineSet *unstructured.Unstructured,
```

Better yet, drop this argument altogether because all callers even send is an empty slice.

This will simplify the function further.

mkolesnik (changes requested):

pkg/ocp/machinesets.go
skitt: Indeed, thanks! I’ve added that as a second commit, so we have the variant with the node list in history if we ever need it.

mkolesnik: Sure makes sense from a commit history perspective :eyes:

mkolesnik (approved):

PR #599, opened 2023-02-16
Remove all "gateway tagging" code

mkolesnik:
This mode was never thoroughly tested, nor was it thoroughly thought through.

When deploying on the cloud, if dedicated gateways are requested then there's sense to use cloud prepare directly on the cloud, outside the K8s sphere of influence.

However, when working with existing nodes its best not to work under K8s and try to change existing entities which are governed by K8s or other systems.

If users desire to run without dedicated gateways, we encourage the use of the Load Balancer mode which is much more K8s oriented and operates through proper K8s interfaces.

Depends on https://github.com/submariner-io/releases/issues/644


dfarrell07:
Do we want to keep this open?

PR #610, opened 2023-02-28
Label gw nodes as infra

### LINKED ISSUE #607: Gateway nodes should be labeled as infrastructure
When we deploy gateway nodes, we should make sure they're labeled as infrastructure nodes, so they don't count against customers' subscriptions.

See https://docs.openshift.com/container-platform/4.12/nodes/nodes/nodes-nodes-creating-infrastructure-nodes.html#creating-an-infra-node_creating-infrastructure-nodes; the relevant label is node-role.kubernetes.io/infra="".

vthapar:
Add `node-role.kubernetes.io/infra=""` label to
gateway nodes.

Fixes #607

### DIFF
--- a/pkg/aws/gw-machineset.go
+++ b/pkg/aws/gw-machineset.go
@@ -42,6 +42,7 @@ spec:
       metadata:
         labels:
           submariner.io/gateway: "true"
+          node-role.kubernetes.io/infra: ""
       taints:
         - effect: NoSchedule
           key: node-role.submariner.io/gateway
--- a/pkg/aws/gw-machineset.yaml.template
+++ b/pkg/aws/gw-machineset.yaml.template
@@ -23,6 +23,7 @@ spec:
       metadata:
         labels:
           submariner.io/gateway: "true"
+          node-role.kubernetes.io/infra: ""
       taints:
         - effect: NoSchedule
           key: node-role.submariner.io/gateway
--- a/pkg/azure/gw-machineset.go
+++ b/pkg/azure/gw-machineset.go
@@ -44,6 +44,7 @@ spec:
       metadata:
         labels:
           submariner.io/gateway: "true"
+          node-role.kubernetes.io/infra: ""
       taints:
         - effect: NoSchedule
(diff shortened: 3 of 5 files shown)


pkg/k8s/k8siface.go
```diff
@@ -100,6 +101,7 @@ func (k *k8sIface) AddGWLabelOnNode(nodeName string) error {
 		}
 
 		labels[SubmarinerGatewayLabel] = "true"
+		labels[InfraNodeLabel] = ""
```
mkolesnik: Since this node isnt managed by us, should we really add this label?

pkg/k8s/k8siface.go
```diff
@@ -124,5 +126,6 @@ func (k *k8sIface) RemoveGWLabelFromWorkerNodes() error {
 func (k *k8sIface) RemoveGWLabelFromWorkerNode(node *v1.Node) error {
 	return k.updateLabel(node.Name, func(existing *v1.Node) {
 		delete(existing.Labels, SubmarinerGatewayLabel)
+		delete(existing.Labels, InfraNodeLabel)
```
mkolesnik: Same question here about deleting it

pkg/k8s/k8siface.go
vthapar: Even I wasn't 100% sure of what would be right way to do this.  Alternately, should we install gw only on nodes that have infra label, at least OCP clusters? Any node running gateway should be treated as infra node but since we're selecting an existing node, as you said, we're unlikely to be owner of it.

EDIT: As per discussion on other issues during PR scrub, we don't really need this. This code will likely get deprecated soon. Will remove this code as well as remove label.

skitt (approved):

mkolesnik (approved):

nyechiel:
@vthapar for some reason, jobs are not running here. Have you tried pushing the commit again or closing and re-opening the PR?

sridhargaddam (approved):

mkolesnik:
@vthapar should this be backported to stable branches?

PR #671, opened 2023-04-12
Skip adding internal Security group in machineset

### LINKED ISSUE #667: In Openstack, cloud prepare does not create gateway node with OVN Kubernetes CNI
**What happened**:
In Openstack, cloud prepare does not create gateway node with OVN Kubernetes CNI

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:
Run cloud prepare in a Openshift cluster with OVNKubernetes CNI

**Anything else we need to know?**:
The below error is shown is machineset controller 

```
E0406 15:03:56.893497       1 controller.go:326]  "msg"="Reconciler error" "error"="error creating Openstack instance: error getting security groups: security group asuryanarhos-r7jdj-submariner-internal-sg not found" "controller"="machine-controller" "name"="asuryanarhos-r7jdj-submariner-gw-0-6vdcb" "namespace"="openshift-machine-api" "object"={"name":"asuryanarhos-r7jdj-submariner-gw-0-6vdcb","namespace":"openshift-machine-api"} "reconcileID"="26d333f5-695e-4213-807e-a24caf5a0b2e"
W0406 15:20:57.362743       1 controller.go:382] asuryanarhos-r7jdj-submariner-gw-0-6vdcb: failed to create machine: error creating Openstack instance: error getting security groups: security group asuryanarhos-r7jdj-submariner-internal-sg not found
```

**Environment**: 
- Diagnose information (use `subctl diagnose all`):
- Gather information (use `subctl gather`):
- Cloud provider or hardware configuration:
- Install tools:
- Others:

aswinsuryan:
Skip adding internal SG in machine set if internal SG does not exist. For deployments with CNI like
OVN there is no extra ports opened for intra cluster traffic.

Fixes #667


pkg/rhos/gw-machineset.go
```diff
@@ -66,12 +66,11 @@ spec:
...
                 tags: openshiftClusterID={{.InfraID}}
           securityGroups:
-          - filter: {}
-            name: {{.InfraID}}-worker
-          - filter: {}
-            name: {{.InfraID}}-submariner-gw-sg
-          - filter: {}
-            name: {{.InfraID}}-submariner-internal-sg
+          - name: {{.InfraID}}-worker
+          {{- if .AddSubmarinerInternalSG }}
```
skitt: Wouldn’t it be more accurate to call this `UseSubmarinerInternalSG` here? Does setting *this* field actually result in the SG being added?

pkg/rhos/ocpgwdeployer.go
```diff
@@ -200,7 +203,15 @@ func (d *ocpGatewayDeployer) deployGWNode(gatewayCount int, groupName string,
...
-			err = d.deployDedicatedGWNode(gatewayNodesToDeploy, status)
+			groupName := d.InfraID + internalSecurityGroupSuffix
+
+			isFound, errSG := checkIfSecurityGroupPresent(groupName, computeClient)
+
+			if errSG != nil {
+				return errSG
+			}
+
+			err = d.deployDedicatedGWNode(gatewayNodesToDeploy, isFound, status)
```
skitt, line 214: Given the description of the change, I find this confusing: we check whether the SG is already present, and if it is, we set a flag called `addInternalSG` to `true`. That suggests that if we find the SG, we add it. Is that really the case, or are we asking `deployDedicatedGWNode` to _use_ the SG we’ve found?

skitt (dismissed):

pkg/rhos/gw-machineset.go
```diff
@@ -25,21 +25,21 @@ metadata:
     machine.openshift.io/cluster-api-cluster: {{.InfraID}}
     machine.openshift.io/cluster-api-machine-role: worker
     machine.openshift.io/cluster-api-machine-type: worker
-  name: {{.InfraID}}-submariner-gw-{{.Index}}
+  name: {{.InfraID}}-submariner-gw-{{.UUID}}
```
sridhargaddam, line 28: Also, wont this change be applicable for the other cloud providers?

aswinsuryan: The internal securtiy group is not added in the same way to the VM in all cloud providers , so this is Openstack specific. Moreover we are testing in OVNKubernetes with other cloud providers and it works.

aswinsuryan:
Depends on: https://github.com/submariner-io/subctl/pull/679

pkg/rhos/ocpgwdeployer.go
aswinsuryan: it is the former,  when it is ovn Kubernetes the internalSG does not exist. so We want to add it only if we find one.

pkg/rhos/gw-machineset.go
aswinsuryan: Done.

dfarrell07:
[E2E (globalnet)](https://github.com/submariner-io/subctl/actions/runs/4731956597/jobs/8397507033?pr=679#logs) timed out, re-running.

dfarrell07:
> Depends on: [submariner-io/subctl#679](https://github.com/submariner-io/subctl/pull/679)

That's merged

FYI the bot only checks the description, not comments.

pkg/rhos/ocpgwdeployer.go
skitt: So we’re looking for a security group, and if we find one, we add it? Doesn’t it already exist if we find it? Or are there two different security groups?

pkg/rhos/ocpgwdeployer.go
```diff
@@ -200,7 +203,15 @@ func (d *ocpGatewayDeployer) deployGWNode(gatewayCount int, groupName string,
 		gatewayNodesToDeploy := gatewayCount - numGatewayNodes
```
sridhargaddam, line 203: In Line 159, we are calling the d.msDeployer.List() which lists the Machines which have the "submariner.io/gateway=true" label. However, I do not see that label added to the MachineSet Yaml[1]. Wont we run into issues (meaning more number of Gateway nodes getting spawned than what we desire) when the cloud-prepare APIs are invoked periodically while it takes time for the actual node to show up? 

[1] The MachineSet is not added with the "submariner.io/gateway=true", however the MachineSet.spec.template.spec.labels include the "submariner.io/gateway=true" label which is actually applied on the final Node that gets created as part of MachineSet.

pkg/rhos/ocpgwdeployer.go
```diff
@@ -209,12 +220,14 @@ func (d *ocpGatewayDeployer) deployGWNode(gatewayCount int, groupName string,
...
 
-func (d *ocpGatewayDeployer) deployDedicatedGWNode(gatewayNodesToDeploy int, status reporter.Interface) error {
+func (d *ocpGatewayDeployer) deployDedicatedGWNode(gatewayNodesToDeploy int, addInternalSG bool,
+	status reporter.Interface,
+) error {
 	for i := 0; i < gatewayNodesToDeploy; i++ {
 		gwNodeName := d.InfraID + "-submariner-gw" + strconv.Itoa(i)
 		status.Start("Deploying dedicated Submariner gateway node %s", gwNodeName)
 
-		err := d.deployGateway(strconv.Itoa(i))
```
sridhargaddam, line 217: Using "strconv.Itoa(i)" was probably masking the issue that I pointed to above. But once we change it to use UUID in the Yaml, I feel we will run into the problem more often.

sridhargaddam (changes requested):

pkg/rhos/ocpgwdeployer.go
aswinsuryan: The internal-sg does not exist always. It is created by openPort api called during cloud prepare. For ovn Kubernetes deployment, this SG is not created

Here we are adding it to machineset only if the SG exists else we skip it.

pkg/rhos/ocpgwdeployer.go
aswinsuryan: The API get the list by checking the label inside spec [1] and not the nodes. So this should exist when the machineset is created successfully. So we should be good right?

[1]https://github.com/submariner-io/cloud-prepare/blob/devel/pkg/ocp/machinesets.go#L205

sridhargaddam: Oh yeah, that's a bit confusing unless we carefully look at where the labels are checked. Thanks for clarifying.

sridhargaddam (approved):

pkg/rhos/ocpgwdeployer.go
yboaron: Yeah,  'addInternalSG' is a bit misleading, maybe it will be better to use 'useSubmarinerInternalSG' instead of  'addInternalSG' ?

aswinsuryan: Done.

nyechiel (approved):

PR #672, opened 2023-04-13
Ignore "InvalidGroup.Duplicate" AWS error

tpantelis:
This is returned from `CreateSecurityGroup` if the SG already exists.

### DIFF
--- a/pkg/aws/securitygroups.go
+++ b/pkg/aws/securitygroups.go
@@ -162,7 +162,8 @@ func (ac *awsCloud) createGatewaySG(vpcID string, ports []api.PortSpec) (string,
 		}
 
 		result, err := ac.client.CreateSecurityGroup(context.TODO(), input)
-		if err != nil {
+
+		if err != nil && !isAWSError(err, "InvalidGroup.Duplicate") {
 			return "", errors.Wrap(err, "error creating AWS security group")
 		}
 


dfarrell07 (approved):

mkolesnik (commented):
Sounds reasonable overall, i guess this only happens in some race condition when you somehow create the security group twice (e.g run cloud prepare twice in parallel), though in such a case i expect some other unwanted outcomes which can be less easily avoided.

Do we have a bug reported for this, or is it some preemptive measure?

tpantelis:
> Do we have a bug reported for this, or is it some preemptive measure?

This was seen while testing https://github.com/stolostron/submariner-addon/pull/640. Cloud prepare was initially successful but a subsequent update caused the controller to reconcile again resulting in a degraded condition. [This change](https://github.com/stolostron/submariner-addon/pull/640/commits/20f263ca32829226c34c27969f3ed5f10d5e3b19) fixed it but it's better to ignore such errors here as we do in other cases [here](https://github.com/submariner-io/cloud-prepare/blob/e9a9abf37403590da19cb50656d61db2a2a9a9e6/pkg/aws/securitygroups.go#L74) and [here](https://github.com/submariner-io/cloud-prepare/blob/e9a9abf37403590da19cb50656d61db2a2a9a9e6/pkg/aws/securitygroups.go#L213) to make it idempotent.

skitt (approved):

PR #722, opened 2023-06-09
Add AirGapped flag to GatewayDeployInput

tpantelis:
...and use in **azure** to set `publicIP` to false in the machine set spec as OCP has an admission webhook controller that enforces this setting in an air-gapped deployment.

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -48,6 +48,9 @@ type GatewayDeployInput struct {
 
 	// Use service of type LoadBalancer to deploy Submariner
 	UseLoadBalancer bool
+
+	// Specifies if the underlying deployment is air-gapped.
+	AirGapped bool
 }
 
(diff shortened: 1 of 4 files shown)


skitt (approved):

nyechiel (approved):

pkg/api/api.go
```diff
@@ -48,6 +48,9 @@ type GatewayDeployInput struct {
 
 	// Use service of type LoadBalancer to deploy Submariner
 	UseLoadBalancer bool
+
+	// Specifies if the underlying deployment is air-gapped.
+	AirGapped bool
```
sridhargaddam, line 53: Thanks for addressing this use-case @tpantelis 

In a future release, we may have to do similar change for the other cloud providers. Currently, they do not seem to use an admission controllers for this use-case, so its not urgent.

sridhargaddam (approved):

PR #822, opened 2023-10-25
Remove generic cloud support

tpantelis:
This is no longer used.


sridhargaddam (approved):
@tpantelis ensure that there is no race condition with @skitt ;-)

skitt (approved):

PR #949, opened 2024-06-11
Fix AWS cloud prepare to support OCP 4.16 new naming convention

aswinsuryan:
OCP 4.16 has a new naming convnention, so cloud prepare is
enhanced to support both old and new formats.

### DIFF
--- a/pkg/aws/aws.go
+++ b/pkg/aws/aws.go
@@ -20,6 +20,9 @@ package aws
 
 import (
 	"context"
+	"fmt"
+	"regexp"
+	"strings"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/config"
@@ -38,9 +41,11 @@ const (
 )
 
 type awsCloud struct {
-	client  awsClient.Interface
-	infraID string
-	region  string
+	client               awsClient.Interface
+	infraID              string
+	region               string
+	nodeSGSuffix         string
+	controlPlaneSGSuffix string
 }
 
 // NewCloud creates a new api.Cloud instance which can prepare AWS for Submariner to be deployed on it.
@@ -88,6 +93,37 @@ func DefaultProfile() string {
 	return "default"
 }
 
+func (ac *awsCloud) setSuffixes(vpcID string) error {
+	if ac.nodeSGSuffix != "" {
+		return nil
+	}
+
+	publicSubnets, err := ac.findPublicSubnets(vpcID, ac.filterByName("{infraID}*-public-{region}*"))
+	if err != nil || len(publicSubnets) == 0 {
+		return errors.Wrapf(err, "unable to find the public subnet")
+	}
+
+	pattern := fmt.Sprintf(`%s.*-subnet-public-%s.*`, regexp.QuoteMeta(ac.infraID), regexp.QuoteMeta(ac.region))
+	re := regexp.MustCompile(pattern)
+
+	for i := range publicSubnets {
+		tags := publicSubnets[i].Tags
+		for i := range tags {
+			if strings.Contains(*tags[i].Key, "Name") && re.MatchString(*tags[i].Value) {
+				ac.nodeSGSuffix = "-node"
+				ac.controlPlaneSGSuffix = "-controlplane"
+
+				return nil
+			}
+		}
+	}
+
+	ac.nodeSGSuffix = "-worker-sg"
+	ac.controlPlaneSGSuffix = "-master-sg"
+
+	return nil
+}
+
 func (ac *awsCloud) OpenPorts(ports []api.PortSpec, status reporter.Interface) error {
 	status.Start(messageRetrieveVPCID)
 	defer status.End()
@@ -95,6 +131,11 @@ func (ac *awsCloud) OpenPorts(ports []api.PortSpec, status reporter.Interface) e
 	vpcID, err := ac.getVpcID()
 	if err != nil {
 		return status.Error(err, "unable to retrieve the VPC ID")
+	}
+
+	err = ac.setSuffixes(vpcID)
+	if err != nil {
+		return status.Error(err, "unable to retrieve the security group names")
 	}
 
 	status.Success(messageRetrievedVPCID, vpcID)
@@ -133,6 +174,11 @@ func (ac *awsCloud) ClosePorts(status reporter.Interface) error {
 	vpcID, err := ac.getVpcID()
 	if err != nil {
 		return status.Error(err, "unable to retrieve the VPC ID")
+	}
+
+	err = ac.setSuffixes(vpcID)
+	if err != nil {
+		return status.Error(err, "unable to retrieve the security group names")
 	}
 
 	status.Success(messageRetrievedVPCID, vpcID)
--- a/pkg/aws/aws.go
+++ b/pkg/aws/aws.go
@@ -99,8 +99,12 @@ func (ac *awsCloud) setSuffixes(vpcID string) error {
 	}
 
 	publicSubnets, err := ac.findPublicSubnets(vpcID, ac.filterByName("{infraID}*-public-{region}*"))
-	if err != nil || len(publicSubnets) == 0 {
+	if err != nil {
 		return errors.Wrapf(err, "unable to find the public subnet")
+	}
+
+	if len(publicSubnets) == 0 {
+		return errors.New("no public subnet found")
 	}
 
 	pattern := fmt.Sprintf(`%s.*-subnet-public-%s.*`, regexp.QuoteMeta(ac.infraID), regexp.QuoteMeta(ac.region))
--- a/pkg/aws/ocpgwdeployer.go
+++ b/pkg/aws/ocpgwdeployer.go
@@ -69,9 +69,14 @@ func (d *ocpGatewayDeployer) Deploy(input api.GatewayDeployInput, status reporte
 
 	status.Success(messageRetrievedVPCID, vpcID)
 
+	err = d.aws.setSuffixes(vpcID)
+	if err != nil {
+		return status.Error(err, "unable to retrieve the security group names")
+	}
+
 	status.Start(messageValidatePrerequisites)
 
-	publicSubnets, err := d.aws.findPublicSubnets(vpcID, d.aws.filterByName("{infraID}-public-{region}*"))
+	publicSubnets, err := d.aws.findPublicSubnets(vpcID, d.aws.filterByName("{infraID}*-public-{region}*"))
 	if err != nil {
 		return status.Error(err, "unable to find public subnets")
 	}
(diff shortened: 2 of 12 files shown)


pkg/aws/ocpgwdeployer.go
```diff
@@ -43,6 +45,11 @@ type ocpGatewayDeployer struct {
 
 var PreferredInstances = []string{"c5d.large", "m5n.large"}
 
+var (
+	nodeSGSuffix         = "-worker-sg"
+	controlPlaneSGSuffix = "-master-sg"
```
tpantelis: These are mutable global vars which is inherently unsafe. Perhaps they should be stored in the `awsCloud` instance?

tpantelis (changes requested):

pkg/aws/ocpgwdeployer.go
```diff
@@ -142,6 +151,23 @@ func (d *ocpGatewayDeployer) Deploy(input api.GatewayDeployInput, status reporte
...
 
+func (d *ocpGatewayDeployer) setSubnetSuffixes(publicSubnets []types.Subnet) {
+	pattern := fmt.Sprintf(`%s.*-subnet-public-%s.*`, regexp.QuoteMeta(d.aws.infraID), regexp.QuoteMeta(d.aws.region))
+	re := regexp.MustCompile(pattern)
+
+	for i := range publicSubnets {
+		tags := publicSubnets[i].Tags
+		for i := range tags {
+			if strings.Contains(*tags[i].Key, "Name") {
+				if re.MatchString(*tags[i].Value) {
```
yboaron: maybe we can change it to : 

`if strings.Contains(*tags[i].Key, "Name") && re.MatchString(*tags[i].Value) {

}` ?

aswinsuryan: Done.

pkg/aws/ocpgwdeployer.go
aswinsuryan: Done.

vthapar (changes requested):
Can you add new 4.16 naming convention to the comit msg and description? Makes it easier to understand changes being made and will allow easier reference for future.

pkg/aws/ocpgwdeployer.go
```diff
@@ -51,6 +53,9 @@ func NewOcpGatewayDeployer(cloud api.Cloud, msDeployer ocp.MachineSetDeployer, i
 		return nil, errors.New("the cloud must be AWS")
 	}
 
+	aws.nodeSGSuffix = "-worker-sg"
+	aws.controlPlaneSGSuffix = "-master-sg"
```
yboaron: It makes sense to set default values also under setSubnetSuffixes function

pkg/aws/ocpgwdeployer.go
```diff
@@ -142,6 +149,21 @@ func (d *ocpGatewayDeployer) Deploy(input api.GatewayDeployInput, status reporte
 	return nil
 }
 
+func (d *ocpGatewayDeployer) setSubnetSuffixes(publicSubnets []types.Subnet) {
```
yboaron: If we set here SG suffixes , maybe we should update function name ?

pkg/aws/ocpgwdeployer.go
```diff
@@ -198,18 +220,29 @@ type machineSetConfig struct {
...
+
+	for i := range ownedFilters {
+		result, err = d.aws.client.DescribeInstances(context.TODO(), &ec2.DescribeInstancesInput{
+			Filters: []types.Filter{
+				ec2Filter("vpc-id", vpcID),
+				d.aws.filterByName("{infraID}-worker*"),
+				ownedFilters[i],
+			},
+		})
+		if err != nil {
```
yboaron, line 222: Do we want to return error if we fail to describe instance for single-owned filter or should we log error and continue with the for loop?

pkg/aws/ocpgwdeployer.go
```diff
@@ -142,6 +149,21 @@ func (d *ocpGatewayDeployer) Deploy(input api.GatewayDeployInput, status reporte
...
 
+func (d *ocpGatewayDeployer) setSubnetSuffixes(publicSubnets []types.Subnet) {
+	pattern := fmt.Sprintf(`%s.*-subnet-public-%s.*`, regexp.QuoteMeta(d.aws.infraID), regexp.QuoteMeta(d.aws.region))
+	re := regexp.MustCompile(pattern)
+
+	for i := range publicSubnets {
+		tags := publicSubnets[i].Tags
+		for i := range tags {
+			if strings.Contains(*tags[i].Key, "Name") && re.MatchString(*tags[i].Value) {
+				d.aws.nodeSGSuffix = "-node"
```
yboaron: Maybe it would be useful for debugging to log the calculated SG suffixes

aswinsuryan: Done.

pkg/aws/ocpgwdeployer.go
aswinsuryan: Done

pkg/aws/ocpgwdeployer.go
aswinsuryan: Done.

pkg/aws/ocpgwdeployer.go
aswinsuryan: it could be a real error should we ignore in that case?

pkg/aws/ocpgwdeployer.go
aswinsuryan: Done.

aswinsuryan:
> Can you add new 4.16 naming convention to the comit msg and description? Makes it easier to understand changes being made and will allow easier reference for future.

Done.

pkg/aws/ocpgwdeployer.go
aswinsuryan: Had to remove it since the logic is now moved to initialization. But the names will be printed in other logs.

pkg/aws/aws.go
```diff
@@ -38,28 +41,74 @@ const (
...
+
+	publicSubnets, err := ac.findPublicSubnets(vpcID, ac.filterByName("{infraID}*-public-{region}*"))
+	if err != nil {
+		return err
+	}
+
+	pattern := fmt.Sprintf(`%s.*-subnet-public-%s.*`, regexp.QuoteMeta(ac.infraID), regexp.QuoteMeta(ac.region))
+	re := regexp.MustCompile(pattern)
+
+	for i := range publicSubnets {
```
yboaron: If publicSubnets is empty should we return error?

pkg/aws/aws.go
```diff
@@ -38,28 +41,74 @@ const (
...
+	err := awsCloud.setSuffixes()
+	if err != nil {
+		return nil
+	}
+
+	return awsCloud
+}
+
+func (ac *awsCloud) setSuffixes() error {
+	vpcID, err := ac.getVpcID()
```
yboaron: Should we also log the error  (could be useful for debugging failures) ?

pkg/aws/aws.go
```diff
@@ -38,28 +41,74 @@ const (
...
+
+	pattern := fmt.Sprintf(`%s.*-subnet-public-%s.*`, regexp.QuoteMeta(ac.infraID), regexp.QuoteMeta(ac.region))
+	re := regexp.MustCompile(pattern)
+
+	for i := range publicSubnets {
+		tags := publicSubnets[i].Tags
+		for i := range tags {
+			if strings.Contains(*tags[i].Key, "Name") && re.MatchString(*tags[i].Value) {
+				ac.nodeSGSuffix = "-node"
+				ac.controlPlaneSGSuffix = "-controlplane"
```
yboaron: I assume we can return here

pkg/aws/ocpgwdeployer.go
yboaron: Well, it's corner case, but I guess if it's enough to find a valid instance for a single-owned filter, we can log the error and continue with the for loop.

I assume we should return error here only if we could not find a valid result for any ownedFilter.

pkg/aws/aws.go
```diff
@@ -38,28 +41,74 @@ const (
...
+	awsCloud := &awsCloud{
 		client:  client,
 		infraID: infraID,
 		region:  region,
 	}
+
+	err := awsCloud.setSuffixes()
+	if err != nil {
+		return nil
+	}
```
tpantelis: This function is expected to instantiate and return a non-nil instance. If errors could occur we really should propagate them but that means changing the function signature (the add-on is the only consumer and we could do that). In lieu of that, the other option is to lazily set the suffixes where needed and propagate errors (in `ocpGatewayDeployer.Deploy` and `OpenPorts`).  So in `setSuffixes`, if `nodeSGSuffix` is non-empty, just return.

tpantelis (changes requested):

pkg/aws/ocpgwdeployer.go
aswinsuryan: There is no logger in cloud-prepare now , we can continue and ignore.

pkg/aws/aws.go
aswinsuryan: Done.

pkg/aws/aws.go
aswinsuryan: not relevant now.

pkg/aws/aws.go
aswinsuryan: done.

pkg/aws/aws.go
aswinsuryan: Done.

pkg/aws/aws.go
```diff
@@ -88,6 +93,33 @@ func DefaultProfile() string {
 	return "default"
 }
 
+func (ac *awsCloud) setSuffixes(vpcID string) error {
```
tpantelis, line 96:
```suggestion
func (ac *awsCloud) setSuffixes(vpcID string) error {
    if ac.nodeSGSuffix != "" {
        return nil
    }
```

tpantelis (changes requested):

pkg/aws/aws.go
aswinsuryan: Done.

pkg/aws/aws.go
```diff
@@ -88,6 +93,37 @@ func DefaultProfile() string {
...
+	if ac.nodeSGSuffix != "" {
+		return nil
+	}
+
+	publicSubnets, err := ac.findPublicSubnets(vpcID, ac.filterByName("{infraID}*-public-{region}*"))
+	if err != nil || len(publicSubnets) == 0 {
```
yboaron: Do we want to return nil if err == nil && len(publicSubnets) == 0 ?

pkg/aws/ocpgwdeployer.go
```diff
@@ -198,18 +203,29 @@ type machineSetConfig struct {
...
+		if err != nil {
+			continue
+		}
+
+		if len(result.Reservations) != 0 {
+			break
+		}
 	}
 
 	if len(result.Reservations) == 0 {
```
yboaron: Minor comment here, I think you should also check that result is not NULL (in case d.aws.client.DescribeInstances returns error for all ownedFilters )

pkg/aws/aws.go
aswinsuryan: We should be finding a public subnet else thee rest of the logic may not work.

pkg/aws/ocpgwdeployer.go
aswinsuryan: Done.

yboaron (approved):

tpantelis (approved):

pkg/aws/aws.go
```diff
@@ -88,6 +93,41 @@ func DefaultProfile() string {
...
+		return nil
+	}
+
+	publicSubnets, err := ac.findPublicSubnets(vpcID, ac.filterByName("{infraID}*-public-{region}*"))
+	if err != nil {
+		return errors.Wrapf(err, "unable to find the public subnet")
+	}
+
+	if len(publicSubnets) == 0 {
+		return errors.New("no public subnet found")
```
aswinsuryan, line 107: Moved to a different `if` and if error is `nil`, `errors.wrapf` will not return an error.

vthapar (approved):

PR #1007, opened 2024-09-21
Add custom vpc support in AWS cloud prepare

aswinsuryan:
Add custom vpc support in AWS cloud prepare

### DIFF
--- a/pkg/aws/aws.go
+++ b/pkg/aws/aws.go
@@ -27,6 +27,7 @@ import (
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/service/ec2"
+	"github.com/aws/aws-sdk-go-v2/service/ec2/types"
 	"github.com/pkg/errors"
 	"github.com/submariner-io/admiral/pkg/reporter"
 	"github.com/submariner-io/cloud-prepare/pkg/api"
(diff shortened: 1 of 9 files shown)


out.txt
```diff
@@ -0,0 +1,443 @@
+/*
```
tpantelis: I assume you didn't mean to submit this file.

pkg/aws/aws.go
```diff
@@ -40,36 +41,84 @@ const (
...
+	}
+}
+
 type awsCloud struct {
 	client               awsClient.Interface
 	infraID              string
 	region               string
 	nodeSGSuffix         string
 	controlPlaneSGSuffix string
+	CloudConfig          map[string]interface{}
```
tpantelis: This doesn't need to be public for consistency.

```suggestion
	cloudConfig          map[string]interface{}
```

pkg/aws/ocpgwdeployer.go
```diff
@@ -69,16 +69,35 @@ func (d *ocpGatewayDeployer) Deploy(input api.GatewayDeployInput, status reporte
...
+				publicSubnets = append(publicSubnets, *subnet)
+			}
+		} else {
+			return errors.New("Subnet IDs must be a valid non-empty slice of strings")
+		}
+	} else {
+		publicSubnets, err = d.aws.findPublicSubnets(vpcID, d.aws.filterByName("{infraID}*-public-{region}*"))
+		if err != nil {
+			return status.Error(err, "unable to find public subnets")
+		}
```
tpantelis, line 100: This code block looks the same as the block starting on L152 above except the latter checks for `len(publicSubnets)`. I think we can refactor to a function and let the caller check `len(publicSubnets)`.

pkg/aws/ocpgwdeployer.go
```diff
@@ -97,9 +116,20 @@ func (d *ocpGatewayDeployer) Deploy(input api.GatewayDeployInput, status reporte
 
 	status.Success("Created Submariner gateway security group %s", gatewaySG)
 
+	err = d.processSubnets(vpcID, gatewaySG, publicSubnets, input, status)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
tpantelis:
```suggestion
	return d.processSubnets(vpcID, gatewaySG, publicSubnets, input, status)
```

vthapar (approved):

tpantelis (approved):

PR #1026, opened 2024-10-04
Fix aws cloud prepare failure

aswinsuryan:
Fix aws cloud prepare failure .

When the public subnets are parsed to identify the prefix the variable is redeclared inside the else block and this causes the variable outside scope of the else to be empty after the block exists. Now the variable declaration is removed.


tpantelis (approved):

tpantelis:
> Fix aws cloud prepare failure

Can you be a bit more specific about the failure?

yboaron (approved):

PR #1044, opened 2024-11-12
Restore for loop in aws cloud prepare for owned filters

aswinsuryan:
While retrieving vpc the owned filters should be tried one after the other than together.

### DIFF
--- a/pkg/aws/vpcs.go
+++ b/pkg/aws/vpcs.go
@@ -42,14 +42,20 @@ func (ac *awsCloud) getVpcID() (string, error) {
 	ownedFilters := ac.filterByCurrentCluster()
 	vpcName := ac.withAWSInfo("{infraID}-vpc")
 
-	filters := []types.Filter{
-		ac.filterByName(vpcName),
-	}
-	filters = append(filters, ownedFilters...)
+	for i := range ownedFilters {
+		filters := []types.Filter{
+			ac.filterByName(vpcName),
+			ownedFilters[i],
+		}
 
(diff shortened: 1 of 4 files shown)


vthapar (approved):

tpantelis (approved):

pkg/aws/vpcs.go
```diff
@@ -42,14 +42,20 @@ func (ac *awsCloud) getVpcID() (string, error) {
...
 
-	result, err = ac.client.DescribeVpcs(context.TODO(), &ec2.DescribeVpcsInput{Filters: filters})
-	if err != nil {
-		return "", errors.Wrap(err, "error describing AWS VPCs")
+		result, err = ac.client.DescribeVpcs(context.TODO(), &ec2.DescribeVpcsInput{Filters: filters})
+		if err != nil {
+			return "", errors.Wrap(err, "error describing AWS VPCs")
+		}
+
+		if len(result.Vpcs) != 0 {
```
yboaron, line 56: Do we need to verify that **result** is not nil ? or it is guaranteed that .client.DescribeVpcs will return non nil value ?

yboaron (approved):

pkg/aws/vpcs.go
aswinsuryan: wouldn't that result in an error ?

tpantelis: It's usually assumed that a valid, non-nil value is returned if no error is returned.

PR #1216, opened 2025-11-03
Fix CVE in release-0.21

dfarrell07:
See commit message for details.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/submariner-io/cloud-prepare
 
-go 1.24.0
+go 1.24.9
 
 require (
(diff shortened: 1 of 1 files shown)


tpantelis (approved):

skitt (changes requested):
This defines the language level, not the toolchain. The appropriate fix here is either to add a toolchain specifier in `go.mod`, or to ensure that the build toolchain is 1.24.9 or later (in the 1.24 branch) — and that is already the case, the images provide a version of Go that has the fixes.

dfarrell07:
Discussing best strategy here: https://github.com/submariner-io/submariner/pull/3695

PR #1274, opened 2026-01-29
Consolidate AWS client creation

tpantelis:
There are 2 ways for users to create an AWS client: directly via `client.New` (used by ACM) and indirectly via `aws.NewCloudFromSettings` (used by subctl) that differ in how the credentials are specified. To simplify the interface and avoid duplication, remove `aws.NewCloudFromSettings` and merge it into `client.New`, with the credentials specified as additional parameters.

Also added unit tests for `client.New`.

## Summary by CodeRabbit

* **New Features**
  * Options-based AWS credential/configuration with helpers for direct creds, credential files, and profiles.

* **Refactor**
  * Consolidated client initialization to the new options pattern; legacy helper constructors for AWS config/credentials removed from one provider to simplify initialization.

* **Tests**
  * Added test suite and comprehensive tests validating credential resolution across direct credentials, files, and profiles.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/pkg/aws/client/client.go
+++ b/pkg/aws/client/client.go
@@ -54,10 +54,39 @@ type Interface interface {
 		optFns ...func(*ec2.Options)) (*ec2.RevokeSecurityGroupIngressOutput, error)
 }
 
-func New(ctx context.Context, accessKeyID, secretAccessKey, region string) (Interface, error) {
-	cfg, err := config.LoadDefaultConfig(ctx,
-		config.WithRegion(region),
-		config.WithCredentialsProvider(credentials.NewStaticCredentialsProvider(accessKeyID, secretAccessKey, "")))
+type Option = config.LoadOptionsFunc
+
+func WithCredentials(accessKeyID, secretAccessKey string) Option {
+	return config.WithCredentialsProvider(credentials.NewStaticCredentialsProvider(accessKeyID, secretAccessKey, ""))
+}
+
+func WithCredentialsFile(file string) Option {
+	return config.WithSharedCredentialsFiles([]string{file})
+}
+
+func WithConfigProfile(p string) Option {
+	return config.WithSharedConfigProfile(p)
+}
+
+// DefaultCredentialsFile returns the default credentials file name.
+func DefaultCredentialsFile() string {
+	return config.DefaultSharedCredentialsFilename()
+}
+
+// DefaultProfile returns the default profile name.
+func DefaultProfile() string {
+	return "default"
+}
+
+func New(ctx context.Context, region string, opts ...Option) (Interface, error) {
+	options := make([]func(*config.LoadOptions) error, 0, 1+len(opts))
+	options = append(options, config.WithRegion(region))
+
+	for _, opt := range opts {
+		options = append(options, opt)
+	}
+
+	cfg, err := config.LoadDefaultConfig(ctx, options...)
 	if err != nil {
 		return nil, fmt.Errorf("failed to load AWS configuration: %w", err)
 	}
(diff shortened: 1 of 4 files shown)


aswinsuryan (approved):

pkg/aws/client/client.go
```diff
@@ -54,10 +54,39 @@ type Interface interface {
...
+	return "default"
+}
+
+func New(ctx context.Context, region string, opts ...Option) (Interface, error) {
+	options := make([]func(*config.LoadOptions) error, 0, 1+len(opts))
+	options = append(options, config.WithRegion(region))
+
+	for _, opt := range opts {
+		options = append(options, opt)
+	}
```
skitt: Avoiding the type alias here allows the function to be slightly simpler:


```suggestion
func New(ctx context.Context, region string, opts ...func(*config.LoadOptions) error) (Interface, error) {
	options := make([]func(*config.LoadOptions) error, 0, 1+len(opts))
	options = append(options, config.WithRegion(region))
	options = append(options, opts...)
```

pkg/aws/client/client_test.go
```diff
@@ -0,0 +1,157 @@
...
+)
+
+const (
+	testRegion          = "us-west-2"
+	testAccessKeyID     = "test-access-key"
+	testSecretAccessKey = "test-secret-key"
+	testProfileName     = "test-profile"
+)
+
+var ctx = context.TODO()
```
skitt: Contexts shouldn’t be stored in a global variable.

skitt (changes requested):

pkg/aws/client/client.go
tpantelis: Originally I didn't have the type alias but [coderabbit flagged it](https://github.com/submariner-io/cloud-prepare/pull/1274#discussion_r2741881905) (tho it was wrong about the compile error). Having the type alias avoids an explicit type cast in `WithCredentials` et al. I'm not really sure what's better.

skitt: The suggestion above compiles and passes the tests; the type alias is still used in `WithCredentials()` etc., it’s just dropped in `New` so that the slice type is usable.

tpantelis: This doesn't compile (with or without the type alias):

 ` options = append(options, opts...)`

_Cannot use 'opts' (type []Option) as the type []func(*config.LoadOptions) error_

I think both ways are functionally equivalent but I prefer the alias b/c it doesn't require explicit conversion. I asked for claude's opinion:

**Which is Better Here?**

  **The type alias (=) is the correct choice** for your use case because:

  1. **Compatibility**: You're creating Option values by wrapping config.LoadOptionsFunc
  functions (lines 60, 64, 68). With the alias, the Option() conversions are unnecessary
  (they do nothing). With a named type, those conversions would be required.

  2. **Passthrough semantics**: Your New() function passes these options directly to
  config.LoadDefaultConfig() which expects config.LoadOptionsFunc. The alias makes this
  seamless.

  3. **No extension needed**: You're not adding methods or behaviors to Option - you're just
  providing a convenient local name for the AWS SDK type.

skitt: You need to take the full suggestion, not just the `append` line. I’ve pushed a commit to show exactly what I mean.

skitt: The subtlety here is that while a type and an alias are equivalent, a slice of the original type and a slice of the aliased type aren’t. Ensuring that we manipulate slices of the original type ensures that everything can be handled as a slice; then we can transparently use the alias on individual elements of the slice (including with variadic functions).

skitt (approved):

PR #1396, opened 2026-07-08
Fix CVES in release-0.23

vthapar:
Bump x/crypto, x/net, otel
Ignore low-severity x/sys CVE incompatible with branch


.grype.yaml
```diff
@@ -12,3 +12,8 @@ ignore:
       name: google.golang.org/protobuf
   # Ignored due to Go 1.25 requirement
   - vulnerability: GHSA-hfvc-g4fc-pqhx
+
+  # Update requires Go 1.25.0 in tools/ and would downgrade x/crypto and x/net. Low severity doesn't justify breaking changes on stable branch.
```
skitt: No, bumping wouldn’t downgrade `x/crypto` or `x/net`. Bumping the Go language level itself in `tools` is safe, the build environment supports it and Mockery v3.2.5 works fine with Go 1.25.

We can keep the exception (GO-2026-5024 is irrelevant for Submariner), but please fix the comment at least.

skitt (changes requested):

skitt (approved):

